PR #4, opened 2019-10-02
use gogen to add unit tests for getPreviousKey

mallow111:
Previously we merged PR to implement gerPreviousKey and its supporting function, this PR is to add unit test for it.


pkg/sloop/store/typed/eventcounttable_test.go
```diff
@@ -110,12 +82,12 @@ func Test_EventCount_TestMinAndMaxKeys(t *testing.T) {
 		return nil
 	})
 	assert.Nil(t, err)
-	assert.Equal(t, "/eventcount/001546398000/somekind/somenamespace/somenamea/68510937-4ffc-11e9-8e26-1418775557c8", minKey)
-	assert.Equal(t, "/eventcount/001546398000/somekind/somenamespace/somenamec/68510937-4ffc-11e9-8e26-1418775557c8", maxKey)
+	assert.Equal(t, "/eventcount/001546398000/somekind/somenamespace/somename/68510937-4ffc-11e9-8e26-1418775557c80", minKey)
```
duke-harlan: I'm not seeing how the `0` got appended to the `minKey` - not sure why it is there

pkg/sloop/store/typed/eventcounttable_test.go
```diff
@@ -138,3 +208,26 @@ func (_ *EventCountKey) GetTestKey() string {
 func (_ *EventCountKey) GetTestValue() *ResourceEventCounts {
 	return &ResourceEventCounts{}
 }
+
+func (_ *EventCountKey) SetTestKeys() []string {
+	untyped.TestHookSetPartitionDuration(time.Hour)
+	var keys []string
+
+	//add another key in same partition for testing purpose
```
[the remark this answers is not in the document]
mallow111: no, it is correct, when i == 'a', the key's partition is someTs.Add(time.Hour*time.Duration(0)) which is same as someTs

pkg/sloop/store/typed/eventcounttable_test.go
```diff
@@ -138,3 +208,26 @@ func (_ *EventCountKey) GetTestKey() string {
 func (_ *EventCountKey) GetTestValue() *ResourceEventCounts {
 	return &ResourceEventCounts{}
 }
+
+func (_ *EventCountKey) SetTestKeys() []string {
```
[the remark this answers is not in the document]
mallow111: we use go generate to make sure it automatically created this function for each data type. The calling function of SetTestKeys is created by gogen, so which means, local scope might not be a good idea.

pkg/sloop/store/typed/eventcounttable_test.go
mallow111: Inside of function SetTestKeys, the first key(minKey) is actually 
`keys = append(keys, NewEventCountKey(someTs, someKind, someNamespace, someName, someUid+"0").String())` where you can see the 0 append on the uid.

pkg/sloop/store/typed/eventcounttable_test.go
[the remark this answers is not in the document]
duke-harlan: the comment applies to the block of code below, which appears to add multiple records to multiple partitions, the comment indicates one key being and one partition

pkg/sloop/store/typed/eventcounttable_test.go
```diff
@@ -138,3 +208,26 @@ func (_ *EventCountKey) GetTestKey() string {
...
+		gap++
+	}
+	return keys
+}
+
+func (_ *EventCountKey) SetTestValue() *ResourceEventCounts {
+	untyped.TestHookSetPartitionDuration(time.Hour)
+	expectedResult := &ResourceEventCounts{MapMinToEvents: make(map[int64]*EventCounts)}
+	expectedResult.MapMinToEvents[someMinute] = &EventCounts{MapReasonToCount: make(map[string]int32)}
+	expectedResult.MapMinToEvents[someMinute].MapReasonToCount[someReason] = someCount
```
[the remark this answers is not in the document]
mallow111: This piece is originally from function ` helper_update_eventcount_table`, I decoupled this function to 2 parts, setKey and setValue, since the original unit test all works fine, so I think we can guarantee it should not break things.

The other thing is if you check the unit test here, it all cares about the keys, for the scope of this PR, I think we should be good.

pkg/sloop/store/typed/watchtable_test.go
```diff
@@ -17,11 +16,14 @@ import (
...
 var someTs = time.Date(2019, 1, 2, 3, 4, 5, 6, time.UTC)
+var zeroData time.Time
 
 const someKind = "somekind"
 const someNamespace = "somenamespace"
 const someName = "somename"
-const somePartition = "001546398000"
+const someMinPartition = "001546398000"
+const someMiddlePartition = "001546401600"
+const someMaxPartition = "001546405200"
```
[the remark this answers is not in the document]
mallow111: these const are shared by all tables(watch,resoureSummary,event,watchActivity),I can move `func (_ *WatchTableKey) SetTestKeys() []string` under these const, but if I move `func (_ *WatchTableKey) SetTestKeys() []string `in this file, it will not make sense, right?

pkg/sloop/store/typed/eventcounttablegen_test.go
```diff
@@ -48,5 +48,67 @@ func Test_ResourceEventCountsTable_SetWorks(t *testing.T) {
...
+	assert.Nil(t, err)
+	return b, wt
+}
+
+func Test_ResourceEventCountsTable_GetUniquePartitionList_Success(t *testing.T) {
+	if helper_ResourceEventCounts_ShouldSkip() {
+		return
+	}
 
+	db, wt := helper_update_ResourceEventCountsTable(t, (&EventCountKey{}).SetTestKeys(), (&EventCountKey{}).SetTestValue())
```
duke-harlan: because the test is not looking at the stored value, please use `ResourceEventCounts{}` instead of `(&EventCountKey{}).SetTestValue()`

pkg/sloop/store/typed/eventcounttable_test.go
```diff
@@ -101,7 +73,7 @@ func Test_EventCount_PutThenGet_SameData(t *testing.T) {
 }
 
 func Test_EventCount_TestMinAndMaxKeys(t *testing.T) {
-	db, rt := helper_update_eventcount_table(t)
+	db, rt := helper_update_ResourceEventCountsTable(t, (&EventCountKey{}).SetTestKeys(), (&EventCountKey{}).SetTestValue())
```
duke-harlan: because the test is not looking at the stored value, please use `ResourceEventCounts{}` instead of `(&EventCountKey{}).SetTestValue()`

duke-harlan: this appears to be the case for numerous tests

duke-harlan: depending upon the number of uses left for `SetTestValue()`, perhaps it can be removed in favor of a locally constructed `ResourceEventCount{}`

pkg/sloop/store/typed/eventcounttable_test.go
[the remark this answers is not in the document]
duke-harlan: all use cases of this appear to be from the same folder, perhaps even just this file - please use keep  scope as minimal as possible

pkg/sloop/store/typed/eventcounttable_test.go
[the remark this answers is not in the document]
duke-harlan: If, as you said, the tests are all about keys, then lets eliminate the potential confusion

pkg/sloop/store/typed/eventcounttable_test.go
```diff
@@ -126,8 +98,106 @@ func Test_EventCount_TestGetMinMaxPartitions(t *testing.T) {
 
 	assert.Nil(t, err)
 	assert.True(t, found)
-	assert.Equal(t, untyped.GetPartitionId(someTs), minPartition)
-	assert.Equal(t, untyped.GetPartitionId(someTs), maxPartition)
+	assert.Equal(t, someMinPartition, minPartition)
+	assert.Equal(t, someMaxPartition, maxPartition)
+}
+
+func Test_EventCount_GetPreviousKey_Success(t *testing.T) {
```
duke-harlan: can this test (and others) be moved in to the template for the generated tests? They seem **very** similar

pkg/sloop/store/typed/watchtable_test.go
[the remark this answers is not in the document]
duke-harlan: I agree that `SetTestKeys` is table specific, but the loop(s) inside it are written to generate N partitions, that control should be moved here and the implementation changed to ensure N partitions are generated.

pkg/sloop/store/typed/eventcounttable_test.go
mallow111: I make `SetTestValue` to include only `ResourceEventCount{}`

pkg/sloop/store/typed/eventcounttable_test.go
mallow111: done

pkg/sloop/store/typed/eventcounttablegen_test.go
```diff
@@ -48,5 +48,96 @@ func Test_ResourceEventCountsTable_SetWorks(t *testing.T) {
...
+	var err1 error
+	err := db.View(func(txn badgerwrap.Txn) error {
+		partList, err1 = wt.GetUniquePartitionList(txn)
+		return err1
+	})
+	assert.Nil(t, err)
+	assert.Len(t, partList, 0)
+}
+
+func Test_EventCount_GetPreviousKey_Success(t *testing.T) {
```
duke-harlan: I'm thinking that tests for `GetPreviousKey` should really be reading multiple keys, not just 1. There should be examples of reading keys in reverse sequence inside of a single partition, and then spanning partitions. There should also be tests that skip over a partition that exists, but doesn't contain any of the key.

PR #11, opened 2019-10-03
Use gogen to add RangeReadPerPartition per table

mallow111:
we want to use this method to replace rangeRead in future, unit test will be in a separate PR, this PR is more focus on the concept and logic of function itself.

### DIFF
--- a/pkg/sloop/store/typed/eventcounttable.go
+++ b/pkg/sloop/store/typed/eventcounttable.go
@@ -12,7 +12,6 @@ import (
 	"github.com/dgraph-io/badger"
 	"github.com/salesforce/sloop/pkg/sloop/store/untyped"
 	"github.com/salesforce/sloop/pkg/sloop/store/untyped/badgerwrap"
-	"reflect"
 	"strings"
 	"time"
 )
@@ -53,8 +52,7 @@ func (k *EventCountKey) Parse(key string) error {
 	return nil
 }
 
-//when some filed is empty, we dont want to return /tablename/partition/////
-//instead we just want to return /tablename/partition
+//todo: need to make sure it can work as keyPrefix when some fields are empty
 func (k *EventCountKey) String() string {
 	return fmt.Sprintf("/%v/%v/%v/%v/%v/%v", k.TableName(), k.PartitionId, k.Kind, k.Namespace, k.Name, k.Uid)
 }
@@ -78,15 +76,4 @@ func (t *ResourceEventCountsTable) GetOrDefault(txn badgerwrap.Txn, key string) 
 
 func (k *EventCountKey) SetPartitionId(newPartitionId string) {
 	k.PartitionId = newPartitionId
-}
-
-func (k *EventCountKey) GetKeyPrefixString() string {
-	v := reflect.ValueOf(*k)
-	str := "/" + k.TableName()
-	for i := 0; i < v.NumField(); i++ {
-		if v.Field(i).String() != "" {
-			str += fmt.Sprintf("/%v", v.Field(i).String())
-		}
-	}
-	return str
 }
--- a/pkg/sloop/store/typed/eventcounttable.go
+++ b/pkg/sloop/store/typed/eventcounttable.go
@@ -12,6 +12,7 @@ import (
 	"github.com/dgraph-io/badger"
 	"github.com/salesforce/sloop/pkg/sloop/store/untyped"
 	"github.com/salesforce/sloop/pkg/sloop/store/untyped/badgerwrap"
+	"reflect"
 	"strings"
 	"time"
 )
@@ -52,6 +53,8 @@ func (k *EventCountKey) Parse(key string) error {
 	return nil
 }
 
+//when some filed is empty, we dont want to return /tablename/partition/////
+//instead we just want to return /tablename/partition
 func (k *EventCountKey) String() string {
 	return fmt.Sprintf("/%v/%v/%v/%v/%v/%v", k.TableName(), k.PartitionId, k.Kind, k.Namespace, k.Name, k.Uid)
 }
@@ -75,4 +78,15 @@ func (t *ResourceEventCountsTable) GetOrDefault(txn badgerwrap.Txn, key string) 
 
 func (k *EventCountKey) SetPartitionId(newPartitionId string) {
 	k.PartitionId = newPartitionId
+}
+
+func (k *EventCountKey) GetKeyPrefixString() string {
+	v := reflect.ValueOf(*k)
+	str := "/" + k.TableName()
+	for i := 0; i < v.NumField(); i++ {
+		if v.Field(i).String() != "" {
+			str += fmt.Sprintf("/%v", v.Field(i).String())
+		}
+	}
+	return str
 }
--- a/pkg/sloop/store/typed/eventcounttablegen.go
+++ b/pkg/sloop/store/typed/eventcounttablegen.go
@@ -132,6 +132,7 @@ func (t *ResourceEventCountsTable) GetMinMaxPartitions(txn badgerwrap.Txn) (bool
 	return true, minKey.PartitionId, maxKey.PartitionId
 }
 
+//TODO: will be replaced by GetPartitionsFromTimeRange in future
 func (t *ResourceEventCountsTable) RangeRead(
 	txn badgerwrap.Txn,
 	keyPredicateFn func(string) bool,
@@ -291,15 +292,14 @@ func (t *ResourceEventCountsTable) RangeReadPerPartition(txn badgerwrap.Txn, key
 
 	lastPartition := ""
 	for _, currentPartition := range partitionList {
-
-		// update keyPrefix's partition with current partition
-		keyPrefix.SetPartitionId(currentPartition)
-
 		curPartitionPrefix := tablePrefix + currentPartition + "/"
 		itr.Seek([]byte(curPartitionPrefix))
 		stats.RowsVisitedCount += 1
 
-		if itr.ValidForPrefix([]byte(keyPrefix.String())) {
+		// update keyPrefix's partition with current partition
+		keyPrefix.SetPartitionId(currentPartition)
+		prefixStr := keyPrefix.GetKeyPrefixString()
+		if itr.ValidForPrefix([]byte(prefixStr)) {
 			stats.RowsPassedKeyPredicateCount += 1
 			key := EventCountKey{}
 			err := key.Parse(string(itr.Item().Key()))
--- a/pkg/sloop/store/typed/eventcounttablegen.go
+++ b/pkg/sloop/store/typed/eventcounttablegen.go
@@ -281,36 +281,38 @@ func (t *ResourceEventCountsTable) RangeReadPerPartition(txn badgerwrap.Txn, key
 	before := time.Now()
 
 	partitionList, err := t.GetPartitionsFromTimeRange(txn, startTime, endTime)
+	stats.PartitionCount = len(partitionList)
 	if err != nil {
 		return resources, stats, errors.Wrapf(err, "failed to get partitions from table:%v, from startTime:%v, to endTime:%v", t.tableName, startTime, endTime)
 	}
 
 	tablePrefix := "/" + t.tableName + "/"
-	iterOpt := badger.DefaultIteratorOptions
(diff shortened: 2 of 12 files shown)


pkg/sloop/store/typed/eventcounttablegen.go
```diff
@@ -270,3 +270,71 @@ func (t *ResourceEventCountsTable) getLastMatchingKeyInPartition(txn badgerwrap.
 	}
 	return false, &EventCountKey{}, nil
 }
+
+//todo: add unit tests
+func (t *ResourceEventCountsTable) RangeReadPerPartition(txn badgerwrap.Txn, keyPrefix *EventCountKey,
```
thomashargrove: I don't think we need both types of RangeRead.  We should be able to have only one that does Prefix, Key and Value predicate IMO

pkg/sloop/store/typed/eventcounttablegen.go
```diff
@@ -270,3 +270,71 @@ func (t *ResourceEventCountsTable) getLastMatchingKeyInPartition(txn badgerwrap.
...
+	stats := RangeReadStats{}
+	before := time.Now()
+	partitionList, err := t.GetUniquePartitionList(txn)
+	if err != nil {
+		return resources, stats, errors.Wrapf(err, "failed to get partition list from table:%v", t.tableName)
+	}
+
+	tablePrefix := "/" + t.tableName + "/"
+	iterOpt := badger.DefaultIteratorOptions
+	iterOpt.Prefix = []byte(tablePrefix)
```
thomashargrove: This prefix will optimize what internal badger tables need to be scanned.  But you are using the table prefix which is very large.  If you make a new iterator for each partition + key prefix you can make this prefix much tighter potentially.  I don't know if it will make a real difference.  Maybe worth a quick experiment?

pkg/sloop/store/typed/eventcounttablegen.go
```diff
@@ -270,3 +270,71 @@ func (t *ResourceEventCountsTable) getLastMatchingKeyInPartition(txn badgerwrap.
...
+	lastPartition := ""
+	for i := len(partitionList) - 1; i >= 0; i-- {
+		currentPartition := partitionList[i]
+		if currentPartition > endPartition {
+			continue
+		} else if currentPartition < startPartition {
+			break
+		} else {
+			curPartitionPrefix := tablePrefix + currentPartition + "/"
+			itr.Seek([]byte(curPartitionPrefix))
```
thomashargrove: You are seeking to the /table/partition/ which could be well before keyPrefix

pkg/sloop/store/typed/eventcounttablegen.go
```diff
@@ -270,3 +270,71 @@ func (t *ResourceEventCountsTable) getLastMatchingKeyInPartition(txn badgerwrap.
...
+			stats.RowsVisitedCount += 1
+
+			if itr.ValidForPrefix([]byte(keyPrefix.String())) {
```
thomashargrove: Where are you updating the partition in keyPrefix?

mallow111: Good call, need to update KeyPrefix's partition based on current Partition

pkg/sloop/store/typed/eventcounttablegen.go
```diff
@@ -270,3 +270,71 @@ func (t *ResourceEventCountsTable) getLastMatchingKeyInPartition(txn badgerwrap.
...
+	tablePrefix := "/" + t.tableName + "/"
+	iterOpt := badger.DefaultIteratorOptions
+	iterOpt.Prefix = []byte(tablePrefix)
+	itr := txn.NewIterator(iterOpt)
+	defer itr.Close()
+
+	startPartition := untyped.GetPartitionId(startTime)
+	endPartition := untyped.GetPartitionId(endTime)
+	lastPartition := ""
+	for i := len(partitionList) - 1; i >= 0; i-- {
```
thomashargrove: please use a better var name instead of `i`

pkg/sloop/store/typed/eventcounttablegen.go
```diff
@@ -270,3 +270,71 @@ func (t *ResourceEventCountsTable) getLastMatchingKeyInPartition(txn badgerwrap.
...
 }
+
+//todo: add unit tests
+func (t *ResourceEventCountsTable) RangeReadPerPartition(txn badgerwrap.Txn, keyPrefix *EventCountKey,
+	valPredicateFn func(*ResourceEventCounts) bool, startTime time.Time, endTime time.Time) (map[EventCountKey]*ResourceEventCounts, RangeReadStats, error) {
+	resources := map[EventCountKey]*ResourceEventCounts{}
+
+	stats := RangeReadStats{}
+	before := time.Now()
+	partitionList, err := t.GetUniquePartitionList(txn)
```
thomashargrove: should we make a new function that gets unique partitions but takes in start and end time to do the filtering there?

pkg/sloop/store/typed/eventcounttablegen.go
```diff
@@ -270,3 +271,84 @@ func (t *ResourceEventCountsTable) getLastMatchingKeyInPartition(txn badgerwrap.
...
+	stats := RangeReadStats{}
+	before := time.Now()
+
+	partitionList, err := t.GetPartitionsFromTimeRange(txn, startTime, endTime)
+	if err != nil {
+		return resources, stats, errors.Wrapf(err, "failed to get partitions from table:%v, from startTime:%v, to endTime:%v", t.tableName, startTime, endTime)
+	}
+
+	tablePrefix := "/" + t.tableName + "/"
+	iterOpt := badger.DefaultIteratorOptions
```
thomashargrove: From a quick test, it is ~20x faster if we use the keyPrefix in the iterator, which means we need to make an iterator for each partition.  And those are very fast (~75 iterator creations per millisecond on my laptop)

pkg/sloop/store/typed/eventcounttablegen.go
```diff
@@ -270,3 +271,84 @@ func (t *ResourceEventCountsTable) getLastMatchingKeyInPartition(txn badgerwrap.
...
+	itr := txn.NewIterator(iterOpt)
+	defer itr.Close()
+
+	lastPartition := ""
+	for _, currentPartition := range partitionList {
+		curPartitionPrefix := tablePrefix + currentPartition + "/"
+		itr.Seek([]byte(curPartitionPrefix))
```
thomashargrove: We want to seek to the keyPrefix if it has at least /table/partition/, else just use /table/partition/

pkg/sloop/store/typed/eventcounttable.go
```diff
@@ -76,3 +79,14 @@ func (t *ResourceEventCountsTable) GetOrDefault(txn badgerwrap.Txn, key string)
 func (k *EventCountKey) SetPartitionId(newPartitionId string) {
 	k.PartitionId = newPartitionId
 }
+
+func (k *EventCountKey) GetKeyPrefixString() string {
```
thomashargrove: Lets hold on this for now, and go back to use string

thomashargrove: For the purpose of faster event lookup, we only need a tweak to the key for watch table.  And one extra if will do:

```
func (k *WatchTableKey) String() string {
	if k.Name == "" && k.Timestamp.IsZero() {
		return fmt.Sprintf("/%v/%v/%v/%v/", k.TableName(), k.PartitionId, k.Kind, k.Namespace)
	} else if k.Timestamp.IsZero() {
		return fmt.Sprintf("/%v/%v/%v/%v/%v/", k.TableName(), k.PartitionId, k.Kind, k.Namespace, k.Name)
	} else {
		return fmt.Sprintf("/%v/%v/%v/%v/%v/%v", k.TableName(), k.PartitionId, k.Kind, k.Namespace, k.Name, k.Timestamp.UnixNano())
	}
}
```

We can build a better key class in a future PR

pkg/sloop/store/typed/eventcounttablegen.go
```diff
@@ -270,3 +271,86 @@ func (t *ResourceEventCountsTable) getLastMatchingKeyInPartition(txn badgerwrap.
...
+	if err != nil {
+		return resources, stats, errors.Wrapf(err, "failed to get partitions from table:%v, from startTime:%v, to endTime:%v", t.tableName, startTime, endTime)
+	}
+
+	tablePrefix := "/" + t.tableName + "/"
+	for _, currentPartition := range partitionList {
+		var seekStr string
+
+		// when keyPrefix does not have partition info(which implies it may not contain kind,namespace,and etc), we seek from /tableName/currentPartition/
+		if keyPrefix.PartitionId == "" {
```
thomashargrove: Can we instead check if keyPrefix is a nil ptr?

pkg/sloop/store/typed/eventcounttablegen.go
```diff
@@ -270,3 +271,86 @@ func (t *ResourceEventCountsTable) getLastMatchingKeyInPartition(txn badgerwrap.
...
+
+		itr := txn.NewIterator(badger.IteratorOptions{Prefix: []byte(seekStr)})
+		defer itr.Close()
+
+		//in worst case, when seekStr = /table/partition, we need to iterate a key list and return all of them
+		//in most cases, we should only hit one result per partition
+		for itr.Seek([]byte(seekStr)); itr.ValidForPrefix([]byte(seekStr)); itr.Next() {
+			stats.RowsVisitedCount += 1
+			key := EventCountKey{}
+			err := key.Parse(string(itr.Item().Key()))
```
thomashargrove: I was expecting us to copy the key prefix and override the partition.  This seems like it would lose the name/namespace/kind for example

mallow111: the override is already happen in line 298,right? it updates its partition with current partition, and move the pointer to the new seek string, did I miss anything?

pkg/sloop/store/typed/eventcounttablegen.go
mallow111: sure, we can do that.

pkg/sloop/store/typed/eventcounttablegen.go
thomashargrove: Why do we need to parse a key if we already have one?

thomashargrove: Sorry, this was my mistake.  I mis-read the code.  Ignore all my comments about this

pkg/sloop/store/typed/eventcounttablegen.go
```diff
@@ -270,3 +271,86 @@ func (t *ResourceEventCountsTable) getLastMatchingKeyInPartition(txn badgerwrap.
...
+		if keyPrefix.PartitionId == "" {
+			seekStr = tablePrefix + currentPartition + "/"
+		} else {
+			// update keyPrefix with current partition
+			keyPrefix.SetPartitionId(currentPartition)
+			seekStr = keyPrefix.String()
+		}
+
+		itr := txn.NewIterator(badger.IteratorOptions{Prefix: []byte(seekStr)})
+		defer itr.Close()
```
thomashargrove: defer will only execute at the end of this function call.  But there may be downsides to keeping a bunch of old iterators open as we make progress through the partitions.  Since Close() is safe to call more than once, we may want to close at the end of each partition, but keep the defer to close in case we return an error

pkg/sloop/store/typed/eventcounttablegen.go
```diff
@@ -270,3 +271,89 @@ func (t *ResourceEventCountsTable) getLastMatchingKeyInPartition(txn badgerwrap.
...
+		return resources, stats, errors.Wrapf(err, "failed to get partitions from table:%v, from startTime:%v, to endTime:%v", t.tableName, startTime, endTime)
+	}
+
+	tablePrefix := "/" + t.tableName + "/"
+	for _, currentPartition := range partitionList {
+		var seekStr string
+
+		// when keyPrefix does not have such info as kind,namespace,and etc, we seek from /tableName/currentPartition/
+		if keyPrefix == nil {
+			seekStr = tablePrefix + currentPartition + "/"
```
thomashargrove: nit: next PR just do `seekStr = "/" + t.tableName + "/" + currentPartition + "/"`

PR #13, opened 2019-10-04
switch config to 'all or none' for CRD watchers, discover list of CRDs

duke-harlan:
Notes:
- the list of CRDs being watched is static based on the set seen at startup - any new CRDs are not watched
- there is some noise in the logs while watching the CRDs that will require some investigation


pkg/sloop/ingress/kubewatcher.go
```diff
@@ -114,27 +107,52 @@ func (i *kubeWatcherImpl) startCustomInformers(requested []CrdWatch) error {
...
+		resource, _ := schema.ParseResourceArg(crd)
 		informer := f.ForResource(*resource)
 		informer.Informer().AddEventHandler(i.getEventHandlerForResource(resource.Resource))
 
-		glog.Infof("Starting customer informer for: %s  [namespace: %s]", watch.ResourceArg, watch.Namespace)
-		go informer.Informer().Run(i.stopChan)
+		go func() {
+			glog.V(2).Infof("Starting CRD informer for: %s", resource)
+			informer.Informer().Run(i.stopChan)
+			glog.V(2).Infof("Exited CRD informer for: %s", resource)
```
thomashargrove: This should never happen right?  If so, can we make it blog.Errorf?

PR #20, opened 2019-10-09
clean up Makefile of sloop

tariq1890:


Makefile
```diff
@@ -1,6 +1,4 @@
-.PHONY:perf perfasm
-
-export GO111MODULE=on
```
tariq1890: Not required in Go 1.13 since it's on by default.

Issue #19, opened 2019-10-08
Publish Helm chart for Sloop


mallow111:
Helm helps to manage Kubernetes applications, Helm Charts help to define, install, and upgrade even the most complex Kubernetes application.

Helm is a package manager (similar to yum and apt), Charts are packages (similar to debs and rpms). For now contributors can just create helm chart under sloop, eventually, we want to contribute helm chart under its official repo: https://github.com/helm/charts

For whoever works on this item, please refer helm chart official doc before starting: https://github.com/helm/helm/tree/master/docs

mallow111:
https://github.com/salesforce/sloop/pull/21

PR #21, opened 2019-10-09
Add helm chart and instruction to use

mallow111:
This is to fix issue https://github.com/salesforce/sloop/issues/19


sloop-helm-charts/README.md
```diff
@@ -0,0 +1,9 @@
+# Use sloop-helm-charts to deploy your service
+1. Build image and tag, and replace the current ones in values.yaml, we are working an image for people to use, it will be release in a few days.
+2. Create a namespace in your cluster for sloop to run, for example: `kubectl create namespace sloop `
```
thomashargrove: nit: * only if you dont already have a namespace

sloop-helm-charts/README.md
```diff
@@ -0,0 +1,9 @@
+# Use sloop-helm-charts to deploy your service
```
thomashargrove: This is instructions for folks not using tiller (or Spinnaker)

sloop-helm-charts/README.md
```diff
@@ -0,0 +1,9 @@
...
+1. Build image and tag, and replace the current ones in values.yaml, we are working an image for people to use, it will be release in a few days.
+2. Create a namespace in your cluster for sloop to run, for example: `kubectl create namespace sloop `
+3. Validate your helm chart in local to make sure there is no mistake:`helm template .`
+4. Write to yamil file: `helm template . --namespace sloop> sloop-test.yaml`
+5. Apply the yaml file in your cluster: `kubectl -n sloop apply -f sloop-test.yaml`
+6.  Check if the service is running: kubectl -n sloop get pods
+7.  Once the pod is running, you can use it by:  `kc-aws port-forward -n sloop statefulset/sloop 8080 8000`
```
thomashargrove: I think this is very rare that folks are running fully locked down clusters, and those folks would already know how to port forward.  At minimum preface this.

thomashargrove:
Dont think we need `sloop-helm-charts/templates/_helpers.tpl` do we?

thomashargrove:
Should we just name the folder `helm`?

sloop-helm-charts/README.md
mallow111: our current helm chart use namespace is default, if people dont change it, they will use default namespace

sloop-helm-charts/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: A Helm chart for Kubernetes
```
jeunii: probably would want to update this and provide more description.

sloop-helm-charts/templates/statefulset.yaml
```diff
@@ -0,0 +1,57 @@
...
+        name: sloop-data
+      - configMap:
+          name: sloop
+        name: sloopconfig
+      serviceAccountName: {{ .Values.serviceAccountName }}
+  volumeClaimTemplates:
+  - metadata:
+     name: sloop-data
+    spec:
+     accessModes: [ "ReadWriteOnce" ]
```
jeunii: Would like to see this come from `values.yaml`.

sloop-helm-charts/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
```
thomashargrove: 0.2

thomashargrove: actually, what is appVersion vs version?

mallow111: appVersion: The version of the app that this contains (optional). This needn't be SemVer.
https://github.com/helm/helm/blob/master/docs/charts.md#the-appversion-field

PR #24, opened 2019-10-10
speed up eventquery

mallow111:
Add both keyPrefix to speed up the eventquery table

### DIFF
--- a/pkg/sloop/queries/eventquery.go
+++ b/pkg/sloop/queries/eventquery.go
@@ -40,7 +40,8 @@ func GetEventData(params url.Values, t typed.Tables, startTime time.Time, endTim
 		selectedNamespace := params.Get(NamespaceParam)
 		selectedName := params.Get(NameParam)
 		selectedKind := params.Get(KindParam)
-		if selectedKind == kubeextractor.NodeKind || selectedKind == kubeextractor.NamespaceKind {
+
+		if kubeextractor.IsClustersScopedResource(selectedKind) {
 			selectedNamespace = DefaultNamespace
 		}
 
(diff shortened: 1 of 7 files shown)


pkg/sloop/queries/eventquery.go
```diff
@@ -36,8 +37,24 @@ func GetEventData(params url.Values, t typed.Tables, startTime time.Time, endTim
...
+			PartitionId: "",
+			Kind:        kubeextractor.EventKind,
+			Namespace:   selectedNamespace,
+			Name:        selectedName,
+			Timestamp:   time.Time{},
+		}
+
 		// TODO: In addition to isEventValInTimeRange we need to also crack open the payload and check the involvedObject kind (+namespace, name, uuid)
-		watchEvents, stats, err2 = t.WatchTable().RangeRead(txn, nil, paramEventDataFn(params), isEventValInTimeRange(startTime, endTime), startTime, endTime)
+		watchEvents, stats, err2 = t.WatchTable().RangeRead(txn, key, paramEventDataFn(params), isEventValInTimeRange(startTime, endTime), startTime, endTime)
```
thomashargrove: minor: since we are already doing the name+namespace+kind using prefix, we dont need the duplicate logic in keyPredicateFn.  All we should need there is a predicate that checks the time range (is the Timestamp in the key between query start and end time).

thomashargrove: My only concern with keeping logic in both places is one can drift from the other with future changes, then debugging is going to be unpleasant

pkg/sloop/queries/eventquery.go
```diff
@@ -36,8 +37,24 @@ func GetEventData(params url.Values, t typed.Tables, startTime time.Time, endTim
 	err := t.Db().View(func(txn badgerwrap.Txn) error {
 		var err2 error
 		var stats typed.RangeReadStats
+		selectedNamespace := params.Get(NamespaceParam)
+		selectedName := params.Get(NameParam)
+		selectedKind := params.Get(KindParam)
+		if selectedKind == kubeextractor.NodeKind || selectedKind == kubeextractor.NamespaceKind {
```
thomashargrove: I suspect we have the logic elsewhere, and we are going to add to it.  Node and Namespace are both cluster-scoped resources, but there are also PSP, ClusterRole, ClusterRoleBinding, Webhooks, etc.  Can you make a function in kubeExtractor like isClustersScopedResource() and use that here and other places with the same logic?

mallow111: yeah, we can make a function to do this.

pkg/sloop/queries/eventquery.go
mallow111: I would like to share a couple of info here:
1. The new RangeRead is already taken care of the time range check, a function called `GetPartitionsFromTimeRange(txn, startTime, endTime)` is called by RangeRead
2. `paramEventDataFn` also check payload to make sure payload's name equals to the name passed by params, we can make paramEventDataFn just focus on payload check.

thomashargrove: RangeRead only handles time check for the matching partitions, which can be day or hour long.  

We need to keep the payloadCheck.  In k8s you can have 2 resources with same name (deployment foo and statefulset foo).  They would both have events with the same prefix (/Events/namespace/foo.xxxx)

PR #26, opened 2019-10-10
Implement multi stage docker build

### LINKED ISSUE #25: Enhance Dockerfile to use multi-stage builds
right now the Dockerfile only builds if the binary was built locally. Would be cool to have a Dockerfile with [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/) to handle this.

jonnylangefeld:
this eliminates local build before docker build

fix #25

Signed-off-by: Jonny Langefeld <jonny.langefeld@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,9 +14,7 @@ linux:
 	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go install -ldflags "-s" -installsuffix cgo -v ./pkg/...
 
 docker: linux
-	cp $(GOPATH)/bin/linux_amd64/sloop .
-	docker build -t sloop .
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -14,9 +14,7 @@ linux:
 	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go install -ldflags "-s" -installsuffix cgo -v ./pkg/...
 
 docker: linux
```
thomashargrove: you should no longer need the dependency on the `linux` target right?

thomashargrove:
Tested it locally and it works fine.

PR #27, opened 2019-10-10
Add liveness and readiness probe

lxlxok:
1. Expose the healthz handler.
2. Add liveness and readiness probe in deployment file.

### DIFF
--- a/helm/templates/statefulset.yaml
+++ b/helm/templates/statefulset.yaml
@@ -29,6 +29,28 @@ spec:
         ports:
         - containerPort: {{ .Values.service.targetPort }}
           protocol: TCP
+        {{- if .Values.livenessProbe.enabled }}
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: {{ .Values.service.targetPort }}
+          initialDelaySeconds: {{ .Values.livenessProbe.initialDelaySeconds }}
+          periodSeconds: {{ .Values.livenessProbe.periodSeconds }}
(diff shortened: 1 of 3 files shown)


thomashargrove:
Thanks Xiao!

helm/templates/statefulset.yaml
```diff
@@ -29,6 +29,28 @@ spec:
...
+          initialDelaySeconds: {{ .Values.livenessProbe.initialDelaySeconds }}
+          periodSeconds: {{ .Values.livenessProbe.periodSeconds }}
+          timeoutSeconds: {{ .Values.livenessProbe.timeoutSeconds }}
+          successThreshold: {{ .Values.livenessProbe.successThreshold }}
+          failureThreshold: {{ .Values.livenessProbe.failureThreshold }}
+        {{- end }}
+        {{- if .Values.readinessProbe.enabled }}
+        readinessProbe:
+          httpGet:
+            path: /
```
mallow111: Referred the example here: https://github.com/helm/charts/blob/master/stable/vault-operator/templates/deployment.yaml#L55, here the path should be `/healthz`?

mallow111:
@lxlxok  Thanks for adding livenessProbe and readinessProbe, one minor pick, also you may want to wait an email [`GitHub] @amateurhuman has invited you to join the @salesforce organization` flow in your github email, it will fix the CLA issue.

helm/templates/statefulset.yaml
lxlxok: Hi @mallow111  Thanks for your review. I updated the path from `/` to `/healthz`.

PR #29, opened 2019-10-10
Fix helm README and last lint error

thomashargrove:
`helm lint .` wanted an icon so I added one (placeholder, getting a proper logo soon)
When we moved the readme it broke the image reference.


thomashargrove:
Thanks @lxlxok I fixed that as well.  Please take a look and re-approve

PR #38, opened 2019-10-15
use consistent config across multiple types of clients

duke-harlan:


thomashargrove:
Let's test it out in-cluster first before merging

PR #55, opened 2019-11-09
fix the misplacement of code caused by go gen

mallow111:
This PR is trying to do 2 things:
1. add itr.Valid() check for corner cases
2. move the code from gen to the right place, othrewise every time when run `make generate` some of the functions will be deleted.


mallow111:
@nurland  This is the second PR that I promised to open in the previous code review, please help to take a look at it.

PR #59, opened 2019-11-12
Fix some of the issues with tooltips going crazy

DuncanSmith1126:
Also update the font size

### DIFF
--- a/pkg/sloop/webfiles/sloop_ui.js
+++ b/pkg/sloop/webfiles/sloop_ui.js
@@ -6,6 +6,7 @@
  */
 
 const displayMaxX = document.documentElement.clientWidth;
+const displayMaxY = document.documentElement.clientHeight;
 
 let topAxis, bottomAxis, xAxisScale, yAxisBand, data, theTime, axisTop, axisBottom, smallBarOffset;
 
@@ -167,7 +168,6 @@ function processAndSortResources(result) {
         }).sort(cmpFn);
         return data
     }
-
 }
 
 function compareKind(a, b) {
@@ -196,15 +196,12 @@ function appendAxes(svg) {
 }
 
 function renderTooltip() {
-    tooltip = d3.select(
-        document.createElement("div"))
+    tooltip = d3.select("#tooltip_container")
+        .append("div")
         .call(createTooltip);
-
-    document.querySelector("body").appendChild(tooltip.node());
 }
 
 function bindMouseEvents(svg) {
-
     svg.on("mousemove", function () {
         let [x, y] = d3.mouse(this);
 
@@ -216,9 +213,9 @@ function bindMouseEvents(svg) {
             line.attr("transform", `translate(${x} 0)`);
             theTime = xAxisScale.invert(x);
             if (!detailedToolTipIsVisible) {
-                tooltip
-                    .style("left", d3.event.pageX + "px")
-                    .style("top", d3.event.pageY + 20 + "px")
+                let tooltipX = d3.event.pageX;
+                let tooltipY = d3.event.pageY;
+                positionTooltip(tooltipX, tooltipY);
             }
         }
     });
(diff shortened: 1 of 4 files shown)


pkg/sloop/webfiles/sloop_ui.js
```diff
@@ -6,6 +6,7 @@
  */
 
 const displayMaxX = document.documentElement.clientWidth;
+const displayMaxY = document.documentElement.clientHeight;
```
duke-harlan: do these need to be reset if the window is resized?

pkg/sloop/webfiles/sloop_ui.js
```diff
@@ -32,25 +33,22 @@ let cmpFn = noSortFn;
...
+    let data = processAndSortResources(result);
+    let svg = render(data);
     bindMouseEvents(svg);
     appendAxes(svg);
     renderTooltip();
 });
 
-function render(result) {
-    let data = processAndSortResources(result);
+function render(data) {
```
duke-harlan: no objection - wondering about the **why** of pulling `processAndSortResources` out, I'm not seeing any other change that leverages this

pkg/sloop/webfiles/sloop_ui.js
```diff
@@ -257,6 +260,7 @@ function bindMouseEvents(svg) {
                     detailedToolTipIsVisible = true;
                     tooltip.html(result);
                     evalJSFromHtml(result);
+                    positionTooltip(tooltipX, tooltipY)
```
duke-harlan: is this going to cause a flicker (visible for a short time in original location/size, then move/resize based on additional data)?

pkg/sloop/webfiles/sloop_ui.js
```diff
@@ -242,9 +240,14 @@ function bindMouseEvents(svg) {
                 }
             );
             tooltip.html(resourceBarHtml);
+            positionTooltip(tooltipX, tooltipY);
```
duke-harlan: is this going to cause a flicker (visible for a short time in original location/size, then move/resize based on additional data)?

pkg/sloop/webfiles/sloop_ui.js
duke-harlan: not sure I understand the need for the `else` half of this if

pkg/sloop/webfiles/sloop_ui.js
```diff
@@ -429,5 +401,65 @@ function createTooltip(el) {
...
+        let resourceBarHtml = getResourceBarContent(
+            {
+                title: d.text,
+                kind: d.kind,
+                namespace: d.namespace,
+                time: theTime
+            }
+        );
+        tooltip.html(resourceBarHtml);
+        positionTooltip(tooltipX, tooltipY);
```
duke-harlan: is this going to cause a flicker (visible for a short time in original location/size, then move/resize based on additional data)?

pkg/sloop/webfiles/sloop_ui.js
DuncanSmith1126: Mostly clarity and future use -esp for animations.

pkg/sloop/webfiles/sloop_ui.js
```diff
@@ -406,7 +374,8 @@ function createResourceBar(d) {
         .attr("x", isLabelRight ? sx - 5 : sx + w + 5)
         .attr("fill", "black")
         .style("text-anchor", isLabelRight ? "end" : "start")
-        .style("dominant-baseline", "hanging");
+        .style("dominant-baseline", "hanging")
+        .style("font-size", "14");
```
thomashargrove: Can we use CSS for the font sizes?  If not, can we define the value at the top of this file so its easy to find and tweak?

thomashargrove:
This is a great improvement for viewing resource details on the main page, but now when I click "Open in new Tab" it renders with a strange half-page on the left.  Is there a way to avoid that?

PR #60, opened 2019-11-13
Fix Helm service name and port forwarding

ryanbrainard:
In the Helm `Service`, the `selector` does not match any of the `labels` of the `StatefulSet`, so the service doesn't actually send traffic to the app. This updates the `selector` to what is [here](https://github.com/salesforce/sloop/blob/4f05081ad2e05a3cb984220de98ddb7478eabee6/helm/sloop/templates/statefulset.yaml#L5). I'm not sure if you wanted a more specific selector, but figured this is a good one to start with.

Now that the service works, it can be used for port-forwarding. Not only is this a better practice, but for some reason, I was having protocol errors forwarding directly to the stateful set that went away after I made this change.

I also updated `kc-aws` to `kubectl`. I'm not sure what that was, but probably some local alias of the original author.


mallow111:
Thank you for making the changes, approved!

PR #64, opened 2019-11-14
Add precise time filtering to event and payload views

thomashargrove:
Before the queries for events and payloads on the resource popup were passing in lookback=5m, but there was code that checked for lookback<30 and changed it to maxLookBack (usually 2 weeks).

This PR adds a second way to specify query time with start_time + end_time.  It also changes the minimum lookback to 1m, and if less just bumps it to 1m.

The benefit of this is now when users click on part of a timeline they only see resources and payloads for +/- 15m, which is helpful for long-lived resources where you just want to investigate one bit of time.  In the future we can make the range adjustable for the user.

### DIFF
--- a/pkg/sloop/queries/timerange.go
+++ b/pkg/sloop/queries/timerange.go
@@ -27,12 +27,11 @@ const minLookback = 1 * time.Minute
 //   We first find the endTime.  If we are looking at historic data, we use the end of the last partitions.  If
 //   that is in the future, we use now().  We don't want to always use now() as that would prevent users from looking
 //   at old data as it would get clipped by maxLookBack
-//
 //   StartTime is just endTime - lookback
 //
 // Using "start_time" and "end_time"
 //
-//   This is straighforward.  These are UTC Unix times
+//   This is straight forward.  These are UTC Unix times
 //
 // TODO: If wall clock is in the middle of the newest partition min-max time we can use it
 func computeTimeRange(params url.Values, tables typed.Tables, maxLookBack time.Duration) (time.Time, time.Time, error) {
@@ -93,11 +92,9 @@ func computeTimeRange(params url.Values, tables typed.Tables, maxLookBack time.D
 	return computedStart, computedEnd, nil
 }
 
-// This looks at our store, and if it has data finds the newest partitions, then finds the end time of that
+// This looks at our store, and if it has data finds the newest partition, then finds the end time of that
 // But if that is in the future we return now
 // This bit of logic is needed for queries with a lookback to determine a good end time
-// For normal operation we general end up with endTime==now(), but if we are viewing a data store in the past
-// we want endTime to be the end of the data set
 func getEndOfTime(tables typed.Tables) time.Time {
 	now := time.Now()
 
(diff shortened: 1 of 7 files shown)


pkg/sloop/queries/timerange.go
```diff
@@ -8,60 +8,152 @@
...
+	var computedEnd time.Time
+	var err error
+
+	// Input validations
+	if startTimeVal == "" && endTimeVal == "" && lookBackVal == "" {
+		return time.Time{}, time.Time{}, fmt.Errorf("Time range must be set with either [%v] or both of [%v,%v]", LookbackParam, StartTimeParam, EndTimeParam)
+	}
+	if lookBackVal != "" {
+		if startTimeVal != "" || endTimeVal != "" {
+			return time.Time{}, time.Time{}, fmt.Errorf("When [%v] is set, you can not set either of [%v,%v]", LookbackParam, StartTimeParam, EndTimeParam)
```
nurland: nit: can we add actual values to the error message?

pkg/sloop/queries/timerange.go
```diff
@@ -8,60 +8,152 @@
...
+		}
+	} else {
+		if (startTimeVal == "") != (endTimeVal == "") {
```
nurland: shouldn't this be `if (startTimeVal == "") || (endTimeVal == "")`? We already know that `lookBack` is not set, so both `startTime` and `endTime` must be set, we cannot proceed if neither of them is set.

thomashargrove: will fix

pkg/sloop/queries/timerange.go
```diff
@@ -8,60 +8,152 @@
...
 
-	// If web request specifies a valid lookback use that, else use the config for the store
-	queryDuration := maxLookBack
-	queryLookBack := params.Get(LookbackParam)
-	if queryLookBack != "" {
-		var err error
-		queryDuration, err = time.ParseDuration(queryLookBack)
+	if lookBackVal != "" {
+		computedEnd = endOfTime
+		lookbackRange, err := getTimeRangeFromLookback(lookBackVal)
```
nurland: nit: maybe rename to getDurationFromLookback? Range implies that we expect start and end.

thomashargrove: done

pkg/sloop/queries/timerange.go
```diff
@@ -8,60 +8,152 @@
...
+	if startTimeVal == "" && endTimeVal == "" && lookBackVal == "" {
+		return time.Time{}, time.Time{}, fmt.Errorf("Time range must be set with either [%v] or both of [%v,%v] but all 3 were empty", LookbackParam, StartTimeParam, EndTimeParam)
+	}
+	if lookBackVal != "" {
+		if startTimeVal != "" || endTimeVal != "" {
+			return time.Time{}, time.Time{}, fmt.Errorf("When [%v] is set, you can not set either of [%v,%v].  Got (%v,%v,%v) respectively", LookbackParam, StartTimeParam, EndTimeParam, lookBackVal, startTimeVal, endTimeVal)
+		}
+	} else {
+		if (startTimeVal == "") || (endTimeVal == "") {
+			return time.Time{}, time.Time{}, fmt.Errorf("Either %v and %v both need to be set or neither set.  Got (%v,%v) respectively", StartTimeParam, EndTimeParam, startTimeVal, endTimeVal)
```
nurland: nit: error message needs to be updated (`Both params [%v,%v] must be set`)

thomashargrove: meh, I don't think anyone would be confused by this error

PR #66, opened 2019-11-14
Remove dupe events for details page

thomashargrove:
Problem: We were showing payload links in the detail view even when the update did not change anything in the payload

Solution: Remove the dupe before we return list to UX

### DIFF
--- a/pkg/sloop/queries/respayloadquery.go
+++ b/pkg/sloop/queries/respayloadquery.go
@@ -133,12 +133,12 @@ func removeDupePayloads(payloads []PayloadOuput) []PayloadOuput {
 
 	ret := []PayloadOuput{}
 
-	lastVal := ""
+	lastPayload := ""
 	for _, val := range payloads {
-		if val.Payload != lastVal {
+		if val.Payload != lastPayload {
 			ret = append(ret, val)
 		}
-		lastVal = val.Payload
+		lastPayload = val.Payload
 	}
 
 	return ret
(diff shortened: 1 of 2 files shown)


pkg/sloop/queries/respayloadquery_test.go
```diff
@@ -146,12 +146,39 @@ func Test_GetResPayload_True_PreviousKeyFound(t *testing.T) {
...
+var somePayloadTs = time.Date(2019, 3, 1, 3, 4, 0, 0, time.UTC)
+
+func Test_removeDupePayloads_emptyWorks(t *testing.T) {
+	ret := removeDupePayloads([]PayloadOuput{})
+	assert.Equal(t, []PayloadOuput{}, ret)
+}
+
+func Test_removeDupePayloads_twoUnique_returnsTwo(t *testing.T) {
+	input := []PayloadOuput{
+		{PayLoadTime: somePayloadTs.Add(time.Minute).Unix(), Payload: "abc"},
```
mallow111: s/somePayloadTs.Add(time.Minute).Unix()/somePayloadTs.Add(time.Minute).UnixNano(), payload time is unixnano

pkg/sloop/queries/respayloadquery.go
```diff
@@ -124,14 +123,23 @@ func getPayloadOutputList(watchRes map[typed.WatchTableKey]*typed.KubeWatchResul
...
+}
+
+func removeDupePayloads(payloads []PayloadOuput) []PayloadOuput {
+	sort.Slice(payloads, func(i, j int) bool {
+		return payloads[i].PayLoadTime < payloads[j].PayLoadTime
+	})
+
+	ret := []PayloadOuput{}
+
+	lastVal := ""
```
nurland: nit: `lastPayload`

pkg/sloop/queries/respayloadquery_test.go
thomashargrove: thanks, fixed

PR #68, opened 2019-11-15
Add payload json blob

mallow111:
This PR makes it possible to display selected payload in page, for compare the diff, it will be in next PR.

Screen Shot 2019-11-15 at 11 44 14 AM


node_modules/vue-json-compare/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
nurland: Why do we have node_modules folder? Can we remove it and reference required scripts via script tag?

package-lock.json
```diff
@@ -0,0 +1,11 @@
+{
```
nurland: what this file is for?

pkg/sloop/webfiles/jdd.js
```diff
@@ -0,0 +1,438 @@
+(function($) {
```
nurland: I don't see the usage of this library in this PR, can we remove it? Also when we start using it - can we reference it via html script tag?

pkg/sloop/webfiles/jdd.min.js
```diff
@@ -0,0 +1,24 @@
+(function($){$.fn.jdd=function(blockLeftId,blockRightId){diffs=[];let left=JSON.parse($(`#${blockLeftId}`).text());let right=JSON.parse($(`#${blockRightId}`).text());let config=createConfig();formatAndDecorate(config,left);$(`#${blockLeftId}`).text(config.out);let config2=createConfig();formatAndDecorate(config2,right);$(`#${blockRightId}`).text(config2.out);config.currentPath=[];config2.currentPath=[];diffVal(left,config,right,config2);diffColor(diffs,blockLeftId,blockRightId);console.log(diffs)};let LEFT='left';let RIGHT='right';let EQUALITY='eq';let TYPE='type';let MISSING='missing';let diffs=[];let requestCount=0;function createConfig(){return{out:'',indent:-1,currentPath:[],paths:[],line:1}};function formatAndDecorate(config,data){if(getType(data)==='array'){formatAndDecorateArray(config,data);return}
```
nurland: please remove

PR #69, opened 2019-11-15
Enable auto push of docker images

nurland:
@thomashargrove

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,3 +2,18 @@ language: go
 
 go:
   - 1.13.x
+
+services:
+  - docker
+
+script:
+  - make
+
+before_deploy:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,4 +1,19 @@
...
+  - make
+
+before_deploy:
+  - echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
+
+deploy:
+  provider: script
+  script:  make docker-push
+  on:
+    branch: master
```
duke-harlan: does the lack of trailing EOL cause any problems?

nurland: I think it shouldn't cause any issues. We can merge and see )

PR #73, opened 2019-11-18
Added aws-iam-authenticator to docker and a launch script

hsiddulugari:
* Added aws-iam-authenticator to docker so that Sloop service inside the docker can authenticate for connecting to AWS EKS clusters
* Added a simple script to
** receive AWS EKS cluster name and optionally region as arguments to setup local kube config
** launch Sloop docker with AWS credentials and local kube config
* A short README section to explain how to use Sloop docker with AWS EKS clusters


Dockerfile
```diff
@@ -7,9 +7,14 @@ RUN go mod download
 
 COPY pkg ./pkg
 
+RUN curl -o /sloop/aws-iam-authenticator https://amazon-eks.s3-us-west-2.amazonaws.com/1.14.6/2019-08-22/bin/linux/amd64/aws-iam-authenticator \
```
thomashargrove: Why is this needed in the dockerfile?  In the bash script below you are not calling it.  And I am able to connect to AWS without it from my workstation.

sloop-to-eks.sh
```diff
@@ -0,0 +1,18 @@
...
+
+if [ $# -lt 1 ] || [ "$1" == "help" ]; then
+    echo "$USAGE"
+    exit 0
+fi
+REGION="us-west-2"
+if [ "$2" != "" ]; then
+    REGION=$2
+fi
+aws eks --region $REGION update-kubeconfig --name $1
```
thomashargrove: This is needed only one time, right?

I am a little hesitant to mess with the users kubeconfig.  This somewhat gets into the responsibility of setting up a workstation to talk to a specific vendor, which I am not sure belongs here.

Without update-kubeconfig this script largely becomes one command, in which case a section in the doc might be better.  Curious what others think though.

Dockerfile
hsiddulugari: When I tried Sloop docker without this change, I got the following error:
```
E1118 07:21:46.316905       1 reflector.go:153] pkg/mod/k8s.io/client-go@v0.0.0-20191003035859-a746c2f219b7/tools/cache/reflector.go:105: Failed to list *v1.Pod: Get https://F5483839EA2647AE729EDBFE036C2559.gr7.us-west-2.eks.amazonaws.com/api/v1/pods?limit=500&resourceVersion=0: getting credentials: exec: exec: "aws-iam-authenticator": executable file not found in $PATH
```
So k8s.io/client-go has a dependency on it.  I don't have to call it explicitly.

sloop-to-eks.sh
hsiddulugari: I definitely juggle with a bunch of clusters myself and don't want to run 2 commands each time so I came up with this script.
This script talks about using Sloop with EKS (AWS) so yes it is vendor (AWS) specific.
Anyway, Sloop seems like a really useful tool.  This script attempts to make it easier to get started for AWS users.

I'm not too attached to the changes I made.  They're useful for me and I thought others might find it useful as well.  If not, I can close this PR.

Dockerfile
thomashargrove: My mistake, I tried it bare metal and it worked but I had this installed.

sloop-to-eks.sh
thomashargrove: I can see the value of this, so lets keep it but put it somewhere other than root.  We already have an "other" folder that would probably work for now.  Or you can suggest something that could be use for all providers and would not be confusing.

Just out of curiosity: Why do you use the `update-kubeconfig` command each time?  I am able to have several EKS clusters in my kubeConfig and I just pass in the `--context` argument to sloop when running against one.  And alternatively I could use `kubeconfig config use-context`.  Is `update-kubeconfig` doing something different?

nurland: @hsiddulugari thanks for contributing!
I agree with @thomashargrove - can we move helper script to `other` folder or create a new one (e.g.: `scripts` or `providers`)? It looks like client-go path forward is to remove all provider specific code from auth(https://github.com/kubernetes/client-go/tree/master/plugin/pkg/client/auth) and just use `exec` auth, so we might need to add gcp/azure scripts in the future.

hsiddulugari: @thomashargrove ,
with `update-kubeconfig`, I'm going for simplicity again (all I have to provide is cluster name) but with `use-context`, I have to provide the full context: `arn:aws:eks:<region>:<account-number>:cluster/<cluster-name>`.  Note the extra account number in it even if we were to build the context ourselves using arguments.  I don't see any harm in doing update-kubeconfig each time.  Do you?

@thomashargrove and @nurland , how about `providers/aws`. I can move the script to that directory.

Issue #58, opened 2019-11-12
-disable-kube-watch should not require kube context


ryanbrainard:
Even though `-disable-kube-watch=true` is set, Sloop still tries to create a Kube context and client. 

```
$ $GOPATH/bin/sloop -disable-kube-watch=true -context=dummy

...
I1113 07:32:36.471984   25355 kubeclient.go:20] Creating k8sclient with user-defined config masterURL=, kubeContext=dummy.
E1113 07:32:36.489076   25355 main.go:31] Main exited with error: failed to create kubernetes client: context "dummy" does not exist
```

Note, I set `-context` here to force it to not use my default. As an alternative reproducer, this `KUBECONFIG` can also be overridden and it fails like this:

```
$ export KUBECONFIG=$HOME/.kube/config-dummy

$ $GOPATH/bin/sloop -disable-kube-watch=true

...
I1113 07:36:46.691441   25775 kubeclient.go:20] Creating k8sclient with user-defined config masterURL=, kubeContext=.
W1113 07:36:46.691501   25775 loader.go:223] Config not found: /Users/rbrainard/.kube/config-dummy
E1113 07:36:46.702665   25775 main.go:31] Main exited with error: failed to create kubernetes client: invalid configuration: no configuration has been provided
W1113 07:36:46.760143   25776 loader.go:223] Config not found: /Users/rbrainard/.kube/config-dummy
error: current-context is not set
W1113 07:36:46.865888   25797 loader.go:223] Config not found: /Users/rbrainard/.kube/config-dummy
error: current-context is not set
```

The actual fix/UX here is a little tricky because Sloop uses the context to point to the correct database and display in the UI. Perhaps if `-disable-kube-watch` is set then `-context` must also be set.

Ref: https://github.com/salesforce/sloop/pull/56#issuecomment-552969933

PR #56, opened 2019-11-12
Data Backup Endpoint

ryanbrainard:
## Description

This adds an endpoint to download a backup of the Badger DB. This is a simple HTTP translation of the built-in [online backup](https://github.com/dgraph-io/badger#database-backup) function. It takes an optional `since` param and returns a streaming download of the backup.

## Use Case

This can surely have many different use cases, but for us, we run ephemeral clusters for testing and need to be able to dump out the contents of Sloop before the cluster is torn down.

### Download Backup

*This is the actual feature added. Shown here with `curl`, but also works in browser; however, I didn't actually add a link to the page since it's probably not something you'd want to have most users casually accessing. Note the use of `-OJ` flags which read the `Content-Disposition` header to get the filename like a browser does.*

```
$ curl -OJ http://localhost:8080/data/backup                                                       
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  212M    0  212M    0     0   228M      0 --:--:-- --:--:-- --:--:--  228M
curl: Saved to filename 'sloop-brainard-l-1-restored-0.bak'
```

### Restore from Backup

*Shown just to demonstrate that the backup actually works. Note, you must use a version of the `badger` CLI that is version compatible. I did this by installing from the go mod cache at the same version hash. In the future, I may send a PR to make allow starting directly from a backup using `DB.Load()` to make this step unnessisary.*

```
$ badger restore --backup-file sloop-brainard-l-1-restored-0.bak --dir /tmp/sloop-restore/brainard-l-1
Listening for /debug HTTP requests at port: 8080
badger 2019/11/12 23:18:16 INFO: All 0 tables opened in 0s
badger 2019/11/12 23:18:17 DEBUG: Storing value log head: {Fid:0 Len:194 Offset:222174181}
badger 2019/11/12 23:18:17 INFO: Got compaction priority: {level:0 score:1.73 dropPrefix:[]}
badger 2019/11/12 23:18:17 INFO: Running for level: 0
badger 2019/11/12 23:18:17 DEBUG: LOG Compact. Added 162292 keys. Skipped 0 keys. Iteration took: 73.077138ms
badger 2019/11/12 23:18:17 DEBUG: Discard stats: map[]
badger 2019/11/12 23:18:17 INFO: LOG Compact 0->1, del 1 tables, add 1 tables, took 127.274711ms
badger 2019/11/12 23:18:17 INFO: Compaction for level: 0 DONE
[...]

PR #75, opened 2019-11-18
Only require valid context when watching

ryanbrainard:
This is a fix to https://github.com/salesforce/sloop/issues/58, which contains more details about the problem itself. The problem was originally raised in https://github.com/salesforce/sloop/pull/56#issuecomment-552969933.

As for the fix, this just splits out a new method from `MakeKubernetesClient()` called `GetKubernetesContext()` that just gets the context, but does not create the client. Then it moves the call to `MakeKubernetesClient()` inside of `if !conf.DisableKubeWatcher { ... }` with the resolved context from `GetKubernetesContext()` so the user preference is only resolved against the config file once.

Fixes https://github.com/salesforce/sloop/issues/58

### DIFF
--- a/pkg/sloop/ingress/kubeclient.go
+++ b/pkg/sloop/ingress/kubeclient.go
@@ -54,7 +54,7 @@ func MakeKubernetesClient(masterURL string, kubeContext string) (kubernetes.Inte
 		return nil, err
 	}
 
-	glog.Infof("Created k8sclient with context=%v, masterURL=%v, configFile=%v.", config.Host, clientConfig.ConfigAccess().GetLoadingPrecedence())
+	glog.Infof("Created k8sclient with context=%v, masterURL=%v, configFile=%v.", kubeContext, config.Host, clientConfig.ConfigAccess().GetLoadingPrecedence())
 	return clientset, nil
 }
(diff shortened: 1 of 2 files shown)


thomashargrove:
Thanks @ryanbrainard!  For testing, have you tried all the following? 1) With watch disabled 2) with watch enabled and no kubeConfig 3) watch with a kubeConfig default context from config 4) watch with kubeConfig and a different context specified on the command line

ryanbrainard:
> 1) With watch disabled
Yes, doesn't try to establish a connection at all but still uses default config.
 
> 2) with watch enabled and no kubeConfig 
Yes, it correctly errors with `invalid configuration: no configuration has been provided`

> 3) watch with a kubeConfig default context from config 
Yes, starts normally

> 4) watch with kubeConfig and a different context specified on the command line
Yes, use the one from the command line.

and one more:

5) with disabled watches and context specified on the command line:
No connection is established, but a DB is still created with the name of context specified, and the UI shows the context. This one is helpful for doing a read-only restore (https://github.com/salesforce/sloop/pull/76) of a DB you don't have kubeConfig on your local machine.

PR #76, opened 2019-11-18
Database Restore

ryanbrainard:
As a follow up to the backup endpoint I added in https://github.com/salesforce/sloop/pull/56, this adds a database restore feature. It works by adding the `-restore-database-file` flag to the `sloop` command, and the DB is opened and restored online from the file using the `DB.Load()` function.

I tested with and without both the `-disable-kube-watch` and `-context` flags and it works as expected. It also appears to work file fine when starting the DB multiple times with the flag since I'm guessing it's just overwriting versions that are already there, but it would be good for someone more familiar with this DB to confirm. 

Note, this works better with the fix I added in https://github.com/salesforce/sloop/pull/75, but it is not required and I ensured there is no merge conflict between the two PRs.


pkg/sloop/ingress/dbrestore_test.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/salesforce/sloop/pkg/sloop/store/untyped/badgerwrap"
+)
+
+func TestDatabaseRestore(t *testing.T) {
+	tempFile, err := ioutil.TempFile("/tmp", "sloop-test-*")
```
thomashargrove: We really try not to let unit tests interact with the filesystem.  Using the host fs makes it so you are testing more than just your unit of code, can lead to flakyness, host specific issues and slow test runs.  (Less of an issue in a small project like this, but a big problem in million+ line codebases)  I would consider using afero, or tweaking the code so you can test the part that does not touch the filesystem.

thomashargrove: In this specific case, there is almost no real business logic in DatabaseRestore of significance.  So simply skipping this unit test is fair game unless you feel it is going to catch some regression.

ryanbrainard: I can just drop the test.

ryanbrainard: Done

PR #79, opened 2019-11-18
attempting to repair damage from previous padding change

duke-harlan:


pkg/sloop/webfiles/resource.html
```diff
@@ -167,7 +167,7 @@ <h2>Viewing range {{.ClickTime}} +/- {{.PlusMinusTime}}</h2>
     });
 </script>
 
-<div id="resource_events">
+<div id="resource_events" style="padding:10px">
```
thomashargrove: Can we set this at the body or document scope, and not on all the divs inside?

PR #80, opened 2019-11-18
Fix diff in inner html

thomashargrove:
Problem: The diff functionality does not work in inner-html
Solution: Add the right includes to the index.html, and comments to remind us to keep them in sync

### DIFF
--- a/pkg/sloop/webfiles/index.html
+++ b/pkg/sloop/webfiles/index.html
@@ -104,9 +104,13 @@ For full license text, see LICENSE.txt file in the repo root or https://opensour
             integrity="sha256-VazP97ZCwtekAsvgPBSUwPFKdrwD3unUfSGVYrahUqU="
             crossorigin="anonymous"></script>
 
+    <!-- This is the list of scripts needed by resource.html, because inner html can not use includes -->
     <script src="https://unpkg.com/axios/dist/axios.min.js" async="" type="text/javascript"></script>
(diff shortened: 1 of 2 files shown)


pkg/sloop/webfiles/index.html
```diff
@@ -104,9 +104,13 @@ <h2>Links</h2>
             integrity="sha256-VazP97ZCwtekAsvgPBSUwPFKdrwD3unUfSGVYrahUqU="
             crossorigin="anonymous"></script>
 
+    <!-- This is the list of scripts needed by resource.html, because inner html can not use includes -->
     <script src="https://unpkg.com/axios/dist/axios.min.js" async="" type="text/javascript"></script>
     <script src="https://cdn.jsdelivr.net/npm/vue/dist/vue.js" async="" type="text/javascript"></script>
     <script src="https://unpkg.com/nprogress@0.2.0/nprogress.js" async="" type="text/javascript"></script>
+    <script src="https://cemerick.github.io/jsdifflib/diffview.js" type="text/javascript" ></script>
+    <script src="https://cemerick.github.io/jsdifflib/difflib.js" type="text/javascript" ></script>
```
nurland: nit: let's add `async` attribute as well.

thomashargrove: Im worried that I dont know how to test this if it introduces a race condition.  Let's save this for another PR.  But good suggestion!

PR #92, opened 2019-12-12
Update badger version. Potential fix for batch writes causing OOM Crashes in Application

DuncanSmith1126:
This is currently being tested - here in case anybody else wants to give it a shot.


pkg/sloop/queries/queryd3heatmap.go
```diff
@@ -305,6 +305,21 @@ func eventCountsToOverlayMap(events map[typed.EventCountKey]*typed.ResourceEvent
 	return retMap, nil
 }
 
+func eventCountsToOverlayMapBucketed(events map[typed.EventCountKey]*typed.ResourceEventCounts, bucketWidth time.Duration) (map[typed.ResourceSummaryKey][]Overlay, error) {
```
thomashargrove: @DuncanSmith1126 this seems unrelated

DuncanSmith1126: Op - yup

PR #100, opened 2020-01-30
Fix for Sloop garbage collection when its not able to successfully decrease the size on disk

sana-jawad:
Problem: Disk usage in our cluster grew way beyond the config limit resulting the application to be OOM killed. There were a couple of issues identified and fixed as follows:

1. Event count table data (although it is the smallest table in terms of number of keys and size) was getting added beyond the max look back time.

```
Before Event Count data is added:    MaxLookBack|----- Count: 50 ---|
                                                                   
After Event Count data is added :      *****MaxLookBack|----- Count: 50 ---|
(new data is shown using *) 
```

2. The sprinkling of event count data was using max look back as a boundary resulting in small partitions containing very less event count data spread all across till look back time.

```
Before Event Count data is added:     MaxLookBack|        |MinPartition -----|
                                                                   
After Event Count data is added:      MaxLookBack|********|MinPartition -----|
```

3. Garbage collection was never called for watch activity table.

Apart from these this PR also adds some new metrics.

### DIFF
--- a/pkg/sloop/processing/eventcount.go
+++ b/pkg/sloop/processing/eventcount.go
@@ -48,33 +48,23 @@ func updateEventCountTable(
 		return nil
 	}
 
-	// Truncate long-lived events to latest of (Now.Timestamp - maxLookBack - partitionDuration)
-	// and (Now.Timestamp - oldestPartitionTimeStamp - partitionDuration)
-	// This avoids filling in data that will immediately
-	// be garbage collected, and avoids creating transactions that are too large and fail
+	// Truncate long-lived events to available partitions
+	// This avoids filling in data that will go beyond the current time range
 
-	// Thus it solves the following scenario where new data is shown using *
-	// Before Event Count data is added:    MaxLookBack|----- Count: 50 ---|
-	// After Event Count data is added :    *****MaxLookBack|----- Count: 50 ---|
+	// Default truncate as computedLastTs -1 * PartitionDuration.
+	// Essentially only allowing events in 1 partition by default.
+	truncateTs := computedLastTs.Add(-1 * untyped.GetPartitionDuration())
 
-	truncateTs := time.Now().UTC().Add(-1*maxLookback + untyped.GetPartitionDuration())
-
-	// The sprinkling of event count data uses max look back as a boundary resulting in small partitions containing very
-	// less event count data spread all across till look back time as shown below. Thus, refining the truncate Timestamp to incorporate MinPartition.
-
-	// Before Event Count data is added:     MaxLookBack|        |MinPartition -----|
-	// After Event Count data is added:      MaxLookBack|********|MinPartition -----|
-	if ok, minPartition, err := tables.GetMinPartition(txn); err == nil && ok {
+	// If there is only one partition, use minPartitionStartTime to ensure we receive events
+	// in that partition.
+	// Otherwise add events to all partitions from MinPartitionEndTime to computedTs.
+	// This ensures no events are added to the very last partition which may get garbage collected soon.
+	if ok, minPartition, maxPartition := tables.EventCountTable().GetMinMaxPartitions(txn); ok {
 		if minPartitionStartTime, minPartitionEndTime, err := untyped.GetTimeRangeForPartition(minPartition); err == nil {
-			// Check if minPartitionEndTime is later then truncateTs, use minPartitionEndTime
-			if minPartitionEndTime.After(truncateTs) {
-				truncateTs = minPartitionEndTime
-			}
-
-			// Check if the minPartitionStartTime is fairly recent i.e. in case of tests or new start of sloop,
-			// use minPartitionStartTime as truncateTs so that we do get events in event count table.
-			if minPartitionStartTime.After(time.Now().UTC().Add(-2 * time.Hour)) {
+			if minPartition == maxPartition {
 				truncateTs = minPartitionStartTime
+			} else {
+				truncateTs = minPartitionEndTime
 			}
 		}
 	}
--- a/pkg/sloop/processing/eventcount.go
+++ b/pkg/sloop/processing/eventcount.go
@@ -9,7 +9,6 @@ package processing
 
 import (
 	"github.com/golang/glog"
-	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/timestamp"
 	"github.com/pkg/errors"
 	"github.com/salesforce/sloop/pkg/sloop/kubeextractor"
@@ -49,13 +48,37 @@ func updateEventCountTable(
 		return nil
 	}
 
-	// Truncate long-lived events to (watch.Timestamp - maxLookback).  This avoids filling in data that will immediately
+	// Truncate long-lived events to latest of (Now.Timestamp - maxLookBack - partitionDuration)
+	// and (Now.Timestamp - oldestPartitionTimeStamp - partitionDuration)
+	// This avoids filling in data that will immediately
 	// be garbage collected, and avoids creating transactions that are too large and fail
-	watchTs, err := ptypes.Timestamp(watchRec.Timestamp)
-	if err != nil {
-		return err
+
+	// Thus it solves the following scenario where new data is shown using *
+	// Before Event Count data is added:    MaxLookBack|----- Count: 50 ---|
+	// After Event Count data is added :    *****MaxLookBack|----- Count: 50 ---|
+
+	truncateTs := time.Now().UTC().Add(-1*maxLookback + untyped.GetPartitionDuration())
+
+	// The sprinkling of event count data uses max look back as a boundary resulting in small partitions containing very
+	// less event count data spread all across till look back time as shown below. Thus, refining the truncate Timestamp to incorporate MinPartition.
+
+	// Before Event Count data is added:     MaxLookBack|        |MinPartition -----|
+	// After Event Count data is added:      MaxLookBack|********|MinPartition -----|
+	if ok, minPartition, err := tables.GetMinPartition(txn); err == nil && ok {
+		if minPartitionStartTime, minPartitionEndTime, err := untyped.GetTimeRangeForPartition(minPartition); err == nil {
+			// Check if minPartitionEndTime is later then truncateTs, use minPartitionEndTime
+			if minPartitionEndTime.After(truncateTs) {
+				truncateTs = minPartitionEndTime
+			}
+
+			// Check if the minPartitionStartTime is fairly recent i.e. in case of tests or new start of sloop,
+			// use minPartitionStartTime as truncateTs so that we do get events in event count table.
+			if minPartitionStartTime.After(time.Now().UTC().Add(-2 * time.Hour)) {
+				truncateTs = minPartitionStartTime
+			}
+		}
 	}
-	truncateTs := watchTs.Add(-1 * maxLookback)
+
 	computedFirstTs, computedLastTs, computedCount = adjustForMaxLookback(computedFirstTs, computedLastTs, computedCount, truncateTs)
 
 	eventCountByMinute := spreadOutEvents(computedFirstTs, computedLastTs, computedCount)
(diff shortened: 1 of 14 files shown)


pkg/sloop/store/typed/eventcounttablegen.go
```diff
@@ -132,6 +132,22 @@ func (t *ResourceEventCountsTable) GetMinMaxPartitions(txn badgerwrap.Txn) (bool
 	return true, minKey.PartitionId, maxKey.PartitionId
 }
 
+func (t *ResourceEventCountsTable) GetMinPartitions(txn badgerwrap.Txn) (bool, string) {
```
thomashargrove: minor: It is only returning a singular partition, so it should not end with `s`

pkg/sloop/store/typed/tables.go
```diff
@@ -20,6 +20,7 @@ type Tables interface {
 	WatchActivityTable() *WatchActivityTable
 	Db() badgerwrap.DB
 	GetMinAndMaxPartition() (bool, string, string, error)
+	GetMinPartition(txn badgerwrap.Txn) (bool, string, error)
```
thomashargrove: Its a little inconsistent to have GetMinPartition(txn) and GetMinAndMaxPartitions(no txn), and also to not have a GetMaxParititon(?).  Would it cause problems to have all 3 and have them take in a txn?

pkg/sloop/storemanager/stats.go
```diff
@@ -16,21 +16,27 @@ const vlogExt = ".vlog" // value log data
...
+	metricCleanedBadgerLsmSizeMb     = promauto.NewGauge(prometheus.GaugeOpts{Name: "sloop_garbage_collected_badger_lsmsizemb"})
+	metricCleanedBadgerVLogFileCount = promauto.NewGauge(prometheus.GaugeOpts{Name: "sloop_garbage_collected_badger_vlogfilecount"})
+	metricCleanedBadgerVLogSizeMb    = promauto.NewGauge(prometheus.GaugeOpts{Name: "sloop_garbage_collected_badger_vlogsizemb"})
 )
 
 type storeStats struct {
 	timestamp         time.Time
-	DiskSizeBytes     uint64
-	DiskLsmBytes      uint64
+	DiskSizeBytes     int64
```
thomashargrove: Why is this changing?

pkg/sloop/storemanager/stats.go
```diff
@@ -99,3 +105,32 @@ func emitMetrics(stats *storeStats) {
...
+func getDeltaStats(beforeStats *storeStats, afterStats *storeStats) *storeStats {
+	ret := &storeStats{}
+	ret.LevelToKeyCount = make(map[int]uint64)
+	ret.LevelToTableCount = make(map[int]int)
+
+	for k, v := range beforeStats.LevelToKeyCount {
+		metricCleanedBadgerKeys.WithLabelValues(fmt.Sprintf("%v", k)).Set(float64(v) - float64(afterStats.LevelToKeyCount[k]))
+	}
+	ret.DiskSizeBytes = beforeStats.DiskSizeBytes - afterStats.DiskSizeBytes
+	fmt.Printf("before %6d and after %6d and delta %6d", beforeStats.DiskSizeBytes, afterStats.DiskSizeBytes, ret.DiskSizeBytes)
```
thomashargrove: dont use `fmt.Printf`.  Either remove or switch to glog

pkg/sloop/storemanager/storemanager.go
```diff
@@ -137,14 +147,15 @@ func (sm *StoreManager) Shutdown() {
...
-func (sm *StoreManager) refreshStats() {
+func (sm *StoreManager) refreshStats() *storeStats {
 	// On startup we have 2 routines trying to do this at the same time
 	// If we have fresh results its good enough
 	if sm.stats != nil && time.Since(sm.stats.timestamp) < time.Second {
-		return
+		return sm.stats
 	}
-	sm.stats = generateStats(sm.config.StoreRoot, sm.tables.Db(), sm.fs)
+	sm.stats =  generateStats(sm.config.StoreRoot, sm.tables.Db(), sm.fs)
```
thomashargrove: run `go fmt` or just do a full `make`

pkg/sloop/processing/eventcount.go
```diff
@@ -49,13 +48,37 @@ func updateEventCountTable(
...
 	// be garbage collected, and avoids creating transactions that are too large and fail
-	watchTs, err := ptypes.Timestamp(watchRec.Timestamp)
-	if err != nil {
-		return err
+
+	// Thus it solves the following scenario where new data is shown using *
+	// Before Event Count data is added:    MaxLookBack|----- Count: 50 ---|
+	// After Event Count data is added :    *****MaxLookBack|----- Count: 50 ---|
+
+	truncateTs := time.Now().UTC().Add(-1*maxLookback + untyped.GetPartitionDuration())
```
thomashargrove: I have big reservations about using `time.Now()`.  It makes for difficult unit tests.  And it means if we want to replay loading data later we will get different results.

The previous code used watchTs.  Is there a reason we can't stick with that?

Also, if we are querying the oldest partition do we even need this anymore?

pkg/sloop/processing/eventcount_test.go
```diff
@@ -21,9 +21,9 @@ import (
 )
 
 var (
-	someEventWatchTs       = time.Date(2019, 8, 29, 21, 24, 55, 6, time.UTC)
+	someEventWatchTs       = time.Now().UTC()
```
thomashargrove: Please use deterministic values in unit tests.  Avoid time.Now() as this is going to be a race condition

pkg/sloop/store/typed/eventcounttablegen.go
thomashargrove: Can we add a method GetMaxPartition, then tweak GetMinAndMaxPartitions to just call both of these, so internally we dont have code duplication?

pkg/sloop/store/typed/tables.go
```diff
@@ -28,6 +29,10 @@ type MinMaxPartitionsGetter interface {
 	GetMinMaxPartitions(badgerwrap.Txn) (bool, string, string)
 }
 
+type MinPartitionsGetter interface {
```
thomashargrove: Should we rename the interface above to be PartitionGetter and add the relevant methods there?

pkg/sloop/store/typed/tables.go
```diff
@@ -94,12 +99,35 @@ func (t *tablesImpl) GetMinAndMaxPartition() (bool, string, string, error) {
...
+		return false, "", nil
+	}
+
+	sort.Strings(allPartitions)
+	return true, allPartitions[0], nil
+}
+
 func (t *tablesImpl) GetTableNames() []string {
-	return []string{t.watchTable.tableName, t.resourceSummaryTable.tableName, t.eventCountTable.tableName}
+	return []string{t.watchTable.tableName, t.resourceSummaryTable.tableName, t.eventCountTable.tableName, t.watchActivityTable.tableName}
```
thomashargrove: thanks for catching and fixing this

pkg/sloop/storemanager/stats.go
```diff
@@ -16,21 +16,27 @@ const vlogExt = ".vlog" // value log data
...
-	metricBadgerVLogFileCount = promauto.NewGauge(prometheus.GaugeOpts{Name: "sloop_badger_vlogfilecount"})
-	metricBadgerVLogSizeMb    = promauto.NewGauge(prometheus.GaugeOpts{Name: "sloop_badger_vlogsizemb"})
+	metricStoreSizeOnDiskMb          = promauto.NewGauge(prometheus.GaugeOpts{Name: "sloop_store_sizeondiskmb"})
+	metricBadgerKeys                 = promauto.NewGaugeVec(prometheus.GaugeOpts{Name: "sloop_badger_keys"}, []string{"level"})
+	metricBadgerTables               = promauto.NewGaugeVec(prometheus.GaugeOpts{Name: "sloop_badger_tables"}, []string{"level"})
+	metricBadgerLsmFileCount         = promauto.NewGauge(prometheus.GaugeOpts{Name: "sloop_badger_lsmfilecount"})
+	metricBadgerLsmSizeMb            = promauto.NewGauge(prometheus.GaugeOpts{Name: "sloop_badger_lsmsizemb"})
+	metricBadgerVLogFileCount        = promauto.NewGauge(prometheus.GaugeOpts{Name: "sloop_badger_vlogfilecount"})
+	metricBadgerVLogSizeMb           = promauto.NewGauge(prometheus.GaugeOpts{Name: "sloop_badger_vlogsizemb"})
+	metricCleanedStoreSizeOnDiskMb   = promauto.NewGauge(prometheus.GaugeOpts{Name: "sloop_garbage_collected_sizeondiskmb"})
```
thomashargrove: I am not sure it will be clear to people what `collected` means.  Is this after GC?  Or delta before->after?

pkg/sloop/store/typed/eventcounttablegen.go
sana-jawad: sure

pkg/sloop/store/typed/tables.go
sana-jawad: changed to PartitionGetter

pkg/sloop/storemanager/stats.go
sana-jawad: The delta of size is a signed number and the delta is calculated suing the difference between  value before and after GC. Hence the type of the variable had to be changed.

pkg/sloop/processing/eventcount_test.go
```diff
@@ -37,15 +41,16 @@ var (
     "uid": "somePodUid"
   },
   "reason":"failed",
-  "firstTimestamp": "2019-08-29T21:24:55Z",
-  "lastTimestamp": "2019-08-29T21:27:55Z",
+  "firstTimestamp": "[firstTimestamp]",
```
thomashargrove: Why are these no longer timestamps?

thomashargrove: oh I see you are replacing it below

pkg/sloop/processing/eventcount.go
```diff
@@ -49,13 +48,27 @@ func updateEventCountTable(
...
-	// be garbage collected, and avoids creating transactions that are too large and fail
-	watchTs, err := ptypes.Timestamp(watchRec.Timestamp)
-	if err != nil {
-		return err
+	// Truncate long-lived events to available partitions
+	// This avoids filling in data that will go beyond the current time range
+
+	// Default truncate as computedLastTs -1 * PartitionDuration.
+	// Essentially only allowing events in 1 partition by default.
+	truncateTs := computedLastTs.Add(-1 * untyped.GetPartitionDuration())
```
thomashargrove: Minor: this should only happen for the first event.  Its a little odd that it computes lastTs-1h where everything else is on a partition boundary.  If we want to be more consistent, we could compute the partition start for lastTs and use that - then we consistently truncate on partition boundaries.  But we could also leave it as-is, since its just one event processing.

For the comment we may want to make it more clear that this will only happen for 1 event on a new sloop with no data.

pkg/sloop/processing/eventcount_test.go
```diff
@@ -116,16 +107,93 @@ func Test_EventCountTable_Event(t *testing.T) {
 
 }
 
+func Test_EventCountTable_TruncateEvent(t *testing.T) {
```
thomashargrove: This test is hard to read, as the timestamp names don't tell me much (firstTs, lastTs).  And this feels like a few different tests put into one.

thomashargrove: I might split this into 4 tests:

1. No data, first event creates 1 partition
1. No data, first event creates 2 partitions
1. There is one partition already in the store
1. There are two partitions in the store

This might be easier with hour partitions, and timestamp variable names that clearly tell me what they are (maybe timestamp_h03m02 or something)

PR #105, opened 2020-02-10
Drop prefix optimization: Replaced DropPrefix with deletion of keys. Also added some more metrics.

sana-jawad:
Right now DropPrefix stops all sloop requests for more than a minute. So replaced drop prefix with deletion.
Added prometheus metrics for age of newest and oldest partitions, cleanup performed flag and number of deleted keys

### DIFF
--- a/pkg/sloop/store/untyped/db_utilities.go
+++ b/pkg/sloop/store/untyped/db_utilities.go
@@ -5,54 +5,75 @@ import (
 	"github.com/salesforce/sloop/pkg/sloop/store/untyped/badgerwrap"
 )
 
-var collectSize = 100000
-
-func DropPrefixNoLock(keyPrefix []byte, db badgerwrap.DB) (error, float64) {
-	var err error
-	var allKeys [][]byte
-
-	// getting all the keys to delete that have the given prefix
-	err = db.View(func(txn badgerwrap.Txn) error {
-		iterOpt := badger.DefaultIteratorOptions
-		iterOpt.Prefix = keyPrefix
-		iterOpt.AllVersions = false
-		iterOpt.PrefetchValues = false
-		it := txn.NewIterator(iterOpt)
-		defer it.Close()
-		for it.Seek(keyPrefix); it.ValidForPrefix(keyPrefix); it.Next() {
-			keyToDel := it.Item().KeyCopy(nil)
-			allKeys = append(allKeys, keyToDel)
+func deleteKeys(db badgerwrap.DB, keysForDelete [][]byte) (error, int) {
+	deletedKeysInThisBatch := 0
+	err := db.Update(func(txn badgerwrap.Txn) error {
+		for _, key := range keysForDelete {
+			err := txn.Delete(key)
+			if err != nil {
+				return err
+			}
+			deletedKeysInThisBatch++
 		}
 		return nil
 	})
+
 	if err != nil {
-		return err, 0
+		return err, deletedKeysInThisBatch
 	}
 
-	// deleting the keys collected in batches
-	numOfKeysDeleted := 0
-	var keysThisBatch [][]byte
-	var deletedKeysInThisBatch = 0
-	for idx, thisKey := range allKeys {
-		keysThisBatch = append(keysThisBatch, thisKey)
-		if len(keysThisBatch) > collectSize || idx == len(allKeys)-1 {
-			err := db.Update(func(txn badgerwrap.Txn) error {
-				for _, keyToDel := range keysThisBatch {
-					txn.Delete(keyToDel)
-					deletedKeysInThisBatch++
-				}
-				return nil
-			})
+	return nil, deletedKeysInThisBatch
+}
 
+func DeleteKeysWithPrefix(keyPrefix []byte, db badgerwrap.DB, deletionBatchSize int) (error, float64, float64) {
+	var err error
+	numOfKeysToDelete := 0
+	numOfKeysDeleted := 0
+	keysLeftToDelete := true
+
+	for keysLeftToDelete {
+
+		keysThisBatch := make([][]byte, 0, deletionBatchSize)
+
+		// getting the keys to delete that have the given prefix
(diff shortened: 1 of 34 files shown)


pkg/sloop/store/untyped/db_utilities.go
```diff
@@ -0,0 +1,56 @@
+package untyped
+
+import (
+	"github.com/dgraph-io/badger/v2"
+	"github.com/salesforce/sloop/pkg/sloop/store/untyped/badgerwrap"
+)
+
+var collectSize = 100000
+
+func DropPrefixNoLock(keyPrefix []byte, db badgerwrap.DB) (error, float64) {
```
duke-harlan: you mentioned metrics - is it worthwhile to publish metrics around what this method sees?
number of keys to delete
number of keys deleted
total time to delete
average time to delete

pkg/sloop/store/untyped/partition.go
```diff
@@ -60,9 +60,17 @@ func GetTimeRangeForPartition(partitionId string) (time.Time, time.Time, error)
...
 }
 
-func TestHookSetPartitionDuration(partDuration time.Duration) bool {
+func GetAgeOfPartitionInHours(partitionId string) (float64, error) {
+	timeForPartition, err := GetTimeForPartition(partitionId)
+	if err != nil {
+		return 0, err
+	}
+
+	return float64((time.Now().Sub(timeForPartition)) / 3600000000000), nil
```
duke-harlan: consider applying a name over this constant - it might be good to calculate this value (i.e. minutes * seconds * milliseconds)

pkg/sloop/storemanager/storemanager.go
```diff
@@ -89,15 +94,17 @@ func (sm *StoreManager) gcLoop() {
...
+		metricGcDeletedNumberOfKeys.Set(numOfDeletedKeys)
 		metricGcRunning.Set(0)
 		if err == nil {
 			metricGcSuccessCount.Inc()
 		} else {
 			metricGcFailedCount.Inc()
 		}
 		metricGcLatency.Set(time.Since(before).Seconds())
-		glog.Infof("GC finished in %v with return %q.  Next run in %v", time.Since(before), err, sm.config.Freq)
+		glog.Infof("GC finished in %v with error %v.  Next run in %v", time.Since(before), err, sm.config.Freq)
```
duke-harlan: consider adding quotes around %v for the error - if the error string resolves to en empty string the result would look like: ` with error .  Next run `, implying an error occurred. The quotes would indicate the error message is blank.

pkg/sloop/storemanager/storemanager.go
```diff
@@ -158,42 +165,52 @@ func (sm *StoreManager) refreshStats() *storeStats {
...
 
+	var totalNumOfDeletedKeys float64 = 0
+	var minPartitionAge float64 = 0
 	anyCleanupPerformed := false
 	if cleanUpTimeCondition(minPartition, maxPartition, timeLimit) || cleanUpFileSizeCondition(stats, sizeLimitBytes) {
 		partStart, partEnd, err := untyped.GetTimeRangeForPartition(minPartition)
 		glog.Infof("GC removing partition %q with data from %v to %v (err %v)", minPartition, partStart, partEnd, err)
-		var errMsgs []string
+
+		minPartitionAge, err = untyped.GetAgeOfPartitionInHours(minPartition)
```
duke-harlan: if `err` is not being used, consider replacing with `_`
if there was an error, age value is 0 - is that the desired metric value?

pkg/sloop/storemanager/storemanager.go
```diff
@@ -158,42 +165,52 @@ func (sm *StoreManager) refreshStats() *storeStats {
...
-		return false, fmt.Errorf("failed to get min partition : %s, max partition: %s, err:%v", minPartition, maxPartition, err)
+		return false, 0, fmt.Errorf("failed to get min partition : %s, max partition: %s, err:%v", minPartition, maxPartition, err)
 	}
 	if !ok {
-		return false, nil
+		return false, 0, nil
 	}
 
+	var totalNumOfDeletedKeys float64 = 0
+	var minPartitionAge float64 = 0
```
duke-harlan: scope of `minPartitionAge` seems to large - if moved to smaller scope, `var` statement can be eliminated

pkg/sloop/store/untyped/db_utilities.go
```diff
@@ -0,0 +1,56 @@
+package untyped
+
+import (
+	"github.com/dgraph-io/badger/v2"
+	"github.com/salesforce/sloop/pkg/sloop/store/untyped/badgerwrap"
+)
+
+var collectSize = 100000
```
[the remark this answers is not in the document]
sana-jawad: Update transaction tracks the rows that are touched and hence very large transactions can end up too much of an overhead.
I picked the 100k value based on how other projects on github are using badger update. There is still a chance that we see latency issues arising from it and I plan to monitor and iterate. And hence it may make sense to move this to a config.

pkg/sloop/store/untyped/db_utilities.go
sana-jawad: This is called from DoCleanup (garbage collection code) which emits both time and number of keys deleted metrics. I can add the metric for number of keys to delete.

pkg/sloop/store/untyped/db_utilities.go
```diff
@@ -0,0 +1,79 @@
...
+				return err, float64(numOfKeysDeleted), float64(numOfKeysToDelete)
+			}
+		}
+
+		if len(keysThisBatch) < deletionBatchSize {
+			keysLeftToDelete = false
+		}
+	}
+
+	if err != nil {
```
duke-harlan: I believe this will never happen. The only code (line#60) where `err` gets set to something other then `nil` already has its own check & return. Consider replacing that with a `break` or removing this check

PR #109, opened 2020-02-25
Sloop keys histogram

sana-jawad:
Added histogram for Sloop with table name, partition id, number of keys and size. It would help in debugging data distribution.

image


pkg/sloop/webserver/debug.go
```diff
@@ -162,6 +164,107 @@ func listKeysHandler(tables typed.Tables) http.HandlerFunc {
...
+}
+
+type histogram struct {
+	HistogramMap map[SloopKey]*SloopKeyInfo
+	TotalKeys    int
+	DeletedKeys  int
+}
+
+// returns IsDeletedExpired, TableName, PartitionId, EstimatedSize, error.
+func ParseSloopKey(item badgerwrap.Item) (bool, string, string, int64, error) {
```
duke-harlan: consider making this local scope - I don't see usage outside of this file

pkg/sloop/webserver/debug.go
```diff
@@ -162,6 +164,107 @@ func listKeysHandler(tables typed.Tables) http.HandlerFunc {
...
+	key := item.Key()
+	parts := strings.Split(string(key), "/")
```
duke-harlan: this logic (key split, verify 6 parts, 1st part is empty) exists in 4 other places - consider creating a shared method instead of duplicating the logic again

pkg/sloop/webserver/debug.go
```diff
@@ -162,6 +164,107 @@ func listKeysHandler(tables typed.Tables) http.HandlerFunc {
...
+	}
+
+	if parts[0] != "" {
+		return false, "", "", 0, fmt.Errorf("key should start with /: %x", key)
+	}
+
+	var isDeletedExpired = item.IsDeletedOrExpired()
+	var tableName = parts[1]
+	var partitionId = parts[2]
+	return isDeletedExpired, tableName, partitionId, item.EstimatedSize(), nil
```
duke-harlan: mixed paradigms - consider moving `item.EstimatedSize()` to be in a local variable like the others

pkg/sloop/webserver/debug.go
```diff
@@ -162,6 +164,107 @@ func listKeysHandler(tables typed.Tables) http.HandlerFunc {
...
+}
+
+func histogramHandler(tables typed.Tables) http.HandlerFunc {
+	return func(writer http.ResponseWriter, request *http.Request) {
+
+		prefix := request.URL.Query().Get("prefix")
+		if len(prefix) > 0 {
```
duke-harlan: does it make sense to allow no prefix as a selection? would the result data set be large or slow?

pkg/sloop/webserver/debug.go
```diff
@@ -162,6 +164,107 @@ func listKeysHandler(tables typed.Tables) http.HandlerFunc {
...
+			defer itr.Close()
+
+			totalKeys := 0
+			isDeletedExpiredKeys := 0
+			var m = make(map[SloopKey]*SloopKeyInfo)
+			for itr.Rewind(); itr.Valid(); itr.Next() {
+				item := itr.Item()
+				isDeletedExpiredOrDiscarded, tableName, partitionId, size, err := ParseSloopKey(item)
+				if err != nil {
+					return errors.Wrapf(err, "failed to print information about key: %x",
```
duke-harlan: error message seems wrong - how does `print` apply to this activity?

pkg/sloop/webserver/debug.go
duke-harlan: why pass the whole item instead of just the key?

pkg/sloop/webserver/debug.go
```diff
@@ -162,6 +164,107 @@ func listKeysHandler(tables typed.Tables) http.HandlerFunc {
...
+		var result histogram
+		err := tables.Db().View(func(txn badgerwrap.Txn) error {
+			iterOpt := badger.DefaultIteratorOptions
+			iterOpt.Prefix = []byte(prefix)
+			iterOpt.PrefetchValues = false
+			itr := txn.NewIterator(iterOpt)
+			defer itr.Close()
+
+			totalKeys := 0
+			isDeletedExpiredKeys := 0
```
duke-harlan: name seems odd - `is` implies a bool, but the usage is a count

pkg/sloop/webserver/debug.go
```diff
@@ -162,6 +164,107 @@ func listKeysHandler(tables typed.Tables) http.HandlerFunc {
...
+			var m = make(map[SloopKey]*SloopKeyInfo)
```
duke-harlan: variable name seems terse - is there a better name?

pkg/sloop/webserver/debug.go
```diff
@@ -162,6 +164,107 @@ func listKeysHandler(tables typed.Tables) http.HandlerFunc {
...
+					return errors.Wrapf(err, "failed to print information about key: %x",
+						item.Key())
+				}
+				totalKeys++
+
+				if isDeletedExpiredOrDiscarded == true {
+					isDeletedExpiredKeys++
+				}
+
+				if m[SloopKey{tableName, partitionId}] == nil {
```
duke-harlan: consider creating a local variable to hold the SloopKey instead of building it multiple times

pkg/sloop/webserver/debug.go
```diff
@@ -162,6 +164,107 @@ func listKeysHandler(tables typed.Tables) http.HandlerFunc {
...
+				item := itr.Item()
+				isDeletedExpiredOrDiscarded, tableName, partitionId, size, err := ParseSloopKey(item)
+				if err != nil {
+					return errors.Wrapf(err, "failed to print information about key: %x",
+						item.Key())
+				}
+				totalKeys++
+
+				if isDeletedExpiredOrDiscarded == true {
+					isDeletedExpiredKeys++
```
duke-harlan: should deleted/expired keys be counted per SloopKey?

pkg/sloop/webserver/debug.go
```diff
@@ -162,6 +164,107 @@ func listKeysHandler(tables typed.Tables) http.HandlerFunc {
...
+			iterOpt := badger.DefaultIteratorOptions
+			iterOpt.Prefix = []byte(prefix)
+			iterOpt.PrefetchValues = false
+			itr := txn.NewIterator(iterOpt)
+			defer itr.Close()
+
+			totalKeys := 0
+			isDeletedExpiredKeys := 0
+			var m = make(map[SloopKey]*SloopKeyInfo)
+			for itr.Rewind(); itr.Valid(); itr.Next() {
```
duke-harlan: ouch - we have to walk through every key to collect this

pkg/sloop/webserver/debug.go
```diff
@@ -162,6 +164,107 @@ func listKeysHandler(tables typed.Tables) http.HandlerFunc {
 	}
 }
 
+type SloopKeyInfo struct {
```
thomashargrove: lets make these private

pkg/sloop/webserver/webserver.go
```diff
@@ -40,6 +40,7 @@ import (
 const (
 	debugViewKeyTemplateFile      = "debugviewkey.html"
 	debugListKeysTemplateFile     = "debuglistkeys.html"
+	debugHistogramFile            = "debughistogram.html"
 	debugConfigTemplateFile       = "debugconfig.html"
 	debugTemplateFile             = "debug.html"
```
thomashargrove: can you update debug.html to include a link to this new page?

sana-jawad: Yes its already updated https://github.com/salesforce/sloop/pull/109/files#diff-39bf5ff9e161f28bf9b37e5848d9d739

pkg/sloop/webfiles/debughistogram.html
```diff
@@ -0,0 +1,48 @@
+<!--
+Copyright (c) 2019, salesforce.com, inc.
```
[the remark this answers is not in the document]
sana-jawad: ah good catch :)

pkg/sloop/webserver/debug.go
sana-jawad: yes we would need to cjeck it.

pkg/sloop/webserver/debug.go
sana-jawad: Because we have to check its other parameters too.

pkg/sloop/webserver/debug.go
sana-jawad: yes to analyze the whole data. It would be slow but as this is for debugging purposes only, so in my opinion it would be ok.

pkg/sloop/webserver/debug.go
```diff
@@ -162,6 +164,117 @@ func listKeysHandler(tables typed.Tables) http.HandlerFunc {
...
+}
+
+type histogram struct {
+	HistogramMap map[sloopKey]*sloopKeyInfo
+	TotalKeys    int
+	DeletedKeys  int
+}
+
+// returns IsDeletedExpired, TableName, PartitionId, TotalSize, error.
+func parseSloopKey(item badgerwrap.Item) (bool, string, string, int64, error) {
```
duke-harlan: - I'm not seeing much value in returning `item.IsDeletedOrExpired()` or `item.EstimatedSize()` as part of this function instead of forcing the caller to get it
- by removing them, the method name also fits better
- it might be convenient for this method to return a `sloopKey`

pkg/sloop/webserver/debug.go
```diff
@@ -162,6 +164,117 @@ func listKeysHandler(tables typed.Tables) http.HandlerFunc {
...
+				for itr.Rewind(); itr.Valid(); itr.Next() {
+					item := itr.Item()
+					isDeletedExpiredOrDiscarded, tableName, partitionId, size, err := parseSloopKey(item)
+					if err != nil {
+						return errors.Wrapf(err, "failed to parse information about key: %x",
+							item.Key())
+					}
+					totalKeys++
+
+					if isDeletedExpiredOrDiscarded == true {
```
duke-harlan: the `== true` is redundant

pkg/sloop/webserver/debug.go
```diff
@@ -162,6 +164,117 @@ func listKeysHandler(tables typed.Tables) http.HandlerFunc {
 	}
 }
 
+type sloopKeyInfo struct {
+	MinimumSize int64
+	MaximumSize int64
+	TotalKeys   int64
+	TotalSize   int64
+	AverageSize float64
```
duke-harlan: if all the other sizes are integers, this probably could be as well

pkg/sloop/webserver/debug.go
sana-jawad: https://github.com/salesforce/sloop/blob/5cd57b189c75a3e801b98d64a47aa9fe159fa576/pkg/sloop/webserver/debug.go#L230 isDeletedOrExpired is used here. Similarly size is also used.

duke-harlan: yes - but they are not used inside this method, they don't really have anything to do with parsing the key, and the caller can just as easily call the accessors to get the values.

pkg/sloop/webserver/debug.go
```diff
@@ -162,6 +164,116 @@ func listKeysHandler(tables typed.Tables) http.HandlerFunc {
...
+}
+
+type histogram struct {
+	HistogramMap map[sloopKey]*sloopKeyInfo
+	TotalKeys    int
+	DeletedKeys  int
+}
+
+// returns TableName, PartitionId, error.
+func parseSloopKey(item badgerwrap.Item) (string, string, error) {
```
duke-harlan: this method could be refactored to return `(sloopKey, error)`

sana-jawad: Yes but I want to keep table name and partition in their separate columns. It helps in better analysis of data.

PR #114, opened 2020-04-17
Sloop was not able to reduce the size on disk after GC runs. This also plays a big role in memory consumption.

sana-jawad:
PR includes following changes:
1. Upgraded Badger version.
2. Added flattening at start up time.
3. Fixed the event count spreading issue which resulted in uneven data distribution across partitions.
4. Moved to drop prefix as it yields better space claim.
5. Added feature flag for switching to delete prefix. Also changed the numberfversions to 0 so that delete prefix would reclaim space. https://github.com/dgraph-io/badger/issues/1228
6. Fixed the issue of unclaimed !badger!move prefixes which are never cleaned up. Details: https://github.com/dgraph-io/badger/issues/1288
7. Added support in debugging pages to see internal keys.

Drop Prefix VS Delete Keys
In first 12 hours GC was run using Drop Prefix and rest of 12 hours were run using Delete Prefix 
image

Debug Page showing internal keys info:

image


pkg/sloop/webserver/debug.go
```diff
@@ -123,23 +124,46 @@ func listKeysHandler(tables typed.Tables) http.HandlerFunc {
...
+
 			iterOpt := badger.DefaultIteratorOptions
 			iterOpt.Prefix = []byte(keyPrefix)
+			iterOpt.AllVersions = true
+			iterOpt.InternalAccess = true
 			itr := txn.NewIterator(iterOpt)
 			defer itr.Close()
 
-			for itr.Seek([]byte(keyPrefix)); itr.ValidForPrefix([]byte(keyPrefix)); itr.Next() {
+			for itr.Rewind(); itr.Valid(); itr.Next() {
```
duke-harlan, line 142: unfamiliar with the subtleties of badger - is there much of a difference between previous loop construct and new one? I was initially wondering if the loop is going to be looking at a lot more rows then before.

pkg/sloop/webserver/debug.go
```diff
@@ -123,23 +124,46 @@ func listKeysHandler(tables typed.Tables) http.HandlerFunc {
...
+				totalCount++
 				thisKey := string(itr.Item().Key())
 				if keyRegEx.MatchString(thisKey) {
-					keys = append(keys, thisKey)
-					count += 1
-					if count >= maxRows {
-						glog.Infof("Reached max rows: %v", maxRows)
-						break
+					if table == "internal" {
+						if !strings.HasPrefix(thisKey, "/eventcount") && !strings.HasPrefix(thisKey, "/ressum") && !strings.HasPrefix(thisKey, "/watch") && !strings.HasPrefix(thisKey, "/watchactivity") {
```
duke-harlan, line 147: is there a positive form of this check (avoiding chain of `!HasPrefix` calls)
if "/oscarthegrouch" is provided, do you still want to continue?

pkg/sloop/webserver/debug.go
```diff
@@ -123,23 +124,46 @@ func listKeysHandler(tables typed.Tables) http.HandlerFunc {
...
+							keys = append(keys, thisKey)
+							count += 1
+							totalSize += itr.Item().EstimatedSize()
+							if count >= maxRows {
+								glog.Infof("Reached max rows: %v", maxRows)
```
duke-harlan, line 152: should probably also log the row count

pkg/sloop/webserver/debug.go
```diff
@@ -123,23 +124,46 @@ func listKeysHandler(tables typed.Tables) http.HandlerFunc {
...
+								break
+							}
+						}
+					} else {
+						keys = append(keys, thisKey)
```
duke-harlan, line 157: consider moving the `else` duplicated code (keys, count, & totalSize) to above the `if` condition

pkg/sloop/storemanager/storemanager.go
```diff
@@ -173,14 +173,53 @@ func (sm *StoreManager) refreshStats() *storeStats {
...
+		}
+
+		glog.Infof("Removed number of keys so far: %v ", totalNumOfDeletedKeys)
+		totalNumOfDeletedKeys += int64(numOfDeletedKeysForPrefix)
+		totalNumOfKeysToDelete += int64(numOfKeysToDeleteForPrefix)
+
+	}
+
+	elapsed := time.Since(beforeGCTime)
+	glog.Infof("Deletion/dropPrefix of prefixes took %v:", elapsed)
```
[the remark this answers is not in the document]
sana-jawad: in case the delete keys is disabled, the dropPrefix would happen for the keys, depending on the feature flag.

pkg/sloop/processing/eventcount.go
```diff
@@ -50,26 +50,22 @@ func updateEventCountTable(
...
-	if ok, minPartition, maxPartition := tables.EventCountTable().GetMinMaxPartitions(txn); ok {
-		if minPartitionStartTime, minPartitionEndTime, err := untyped.GetTimeRangeForPartition(minPartition); err == nil {
-			if minPartition == maxPartition {
-				truncateTs = minPartitionStartTime
-			} else {
-				truncateTs = minPartitionEndTime
-			}
-		}
+	_, minPartitionEndTime, err := untyped.GetTimeRangeForPartition(minPartition)
+	if err != nil && minPartition != "" {
```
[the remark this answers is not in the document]
sana-jawad: When there is no minpartition, the function itself just returns the error and does not continue. Empty partitionID is not handled as a special case as per say.

pkg/sloop/webserver/debug.go
sana-jawad: There is a difference that it actually loops over the keys instead of seek which is relatively a faster call. Even though it should work as expected, however unfortunately in the current setup the deleted keys do end up showing in seek where I had spend a lot of time to debug.

For now I am keeping Rewind instead of seek. However it is one of the things in my todo to see why seek is not behaving appropriately.

On the other hand, thanks to your comment, I just noticed that I am not calling ValidForPrefix which is still required. I will update that.

pkg/sloop/storemanager/storemanager.go
```diff
@@ -173,14 +173,53 @@ func (sm *StoreManager) refreshStats() *storeStats {
 	return sm.stats
 }
 
-func doCleanup(tables typed.Tables, timeLimit time.Duration, sizeLimitBytes int, stats *storeStats, deletionBatchSize int, gcThreshold float64) (bool, int64, int64, error) {
+func doCleanup(tables typed.Tables, timeLimit time.Duration, sizeLimitBytes int, stats *storeStats, deletionBatchSize int, gcThreshold float64, enableDeletePrefix bool) (bool, int64, int64, error) {
```
duke-harlan, line 176: I'm not asking as part of this PR, but hopefully one soon to follow

perhaps this is just me...
I'm struggling to understand the big/medium picture of this code (I think I did last set of changes as well). Under the umbrella of maintainability, is it possible to layer this code or abstract aspects of it so the flow of the logic is more easily observed.

pkg/sloop/common/db_utilities.go
```diff
@@ -5,8 +5,8 @@ import (
 	"github.com/salesforce/sloop/pkg/sloop/store/untyped/badgerwrap"
 )
 
-func deleteKeys(db badgerwrap.DB, keysForDelete [][]byte) (error, int) {
-	deletedKeysInThisBatch := 0
+func deleteKeys(db badgerwrap.DB, keysForDelete [][]byte) (error, uint64) {
+	var deletedKeysInThisBatch uint64 = 0
```
mallow111, line 9: any reason we change from int to uint64?

pkg/sloop/storemanager/storemanager.go
```diff
@@ -173,14 +173,53 @@ func (sm *StoreManager) refreshStats() *storeStats {
...
+		beforeDropPrefix := time.Now()
+		glog.Infof("Deleted %d keys", totalNumOfDeletedKeys)
+		// dropping prefix to force compression in case of keys deleted
+		err := tables.Db().DropPrefix([]byte{})
+		glog.Infof("Drop prefix took %v with error: %v", time.Since(beforeDropPrefix), err)
+	}
+	return anyCleanupPerformed, totalNumOfDeletedKeys, totalNumOfKeysToDelete, nil
+}
 
+func getAndSetMinAndMaxPartitions(tables typed.Tables) (bool, string, string) {
```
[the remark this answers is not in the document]
sana-jawad: it sets the metrics. Let me rename the method.

pkg/sloop/storemanager/storemanager.go
sana-jawad: sure

pkg/sloop/storemanager/storemanager.go
```diff
@@ -193,92 +232,105 @@ func doCleanup(tables typed.Tables, timeLimit time.Duration, sizeLimitBytes int,
...
 	glog.Infof("GC removing partition %q with data from %v to %v (err %v)", minPartition, partStart, partEnd, err)
 	var errMessages []string
 	for _, tableName := range tables.GetTableNames() {
 		prefix := fmt.Sprintf("/%s/%s", tableName, minPartition)
 		start := time.Now()
-		err, numOfDeletedKeysforPrefix, numOfKeysToDeleteForPrefix = common.DeleteKeysWithPrefix([]byte(prefix), tables.Db(), deletionBatchSize)
-		metricGcDeletedNumberOfKeysByTable.WithLabelValues(fmt.Sprintf("%v", tableName)).Set(float64(numOfDeletedKeysforPrefix))
+		numberOfKeysToRemove := partitionInfo.TableNameToKeyCountMap[tableName]
+		if enableDeleteKeys {
+			err, numOfDeletedKeysForPrefix, numOfKeysToDeleteForPrefix = common.DeleteKeysWithPrefix(prefix, tables.Db(), deletionBatchSize, numberOfKeysToRemove)
```
mallow111, line 310: when err != nil, we might want to log the error, and print out numOfDeletedKeysForPrefix, numOfKeysToDeleteForPrefix

pkg/sloop/common/db_utilities.go
sana-jawad: the number of keys cannot be negative so changed it uint64 like other places in code

pkg/sloop/common/db_utilities.go
```diff
@@ -57,30 +58,27 @@ func DeleteKeysWithPrefix(keyPrefix []byte, db badgerwrap.DB, deletionBatchSize
...
-		}
 	}
-
 	return nil, numOfKeysDeleted, numOfKeysToDelete
-
 }
 
-func GetTotalKeyCount(db badgerwrap.DB) uint64 {
+// returns the number of keys in DB with given prefix. If prefix is not provided it gives count of all keys
+func GetTotalKeyCount(db badgerwrap.DB, keyPrefix []byte) uint64 {
```
mallow111, line 71: for keyPrefix, I see some use string, some use []byte, we can just use one form and apply it everywhere.

pkg/sloop/processing/eventcount.go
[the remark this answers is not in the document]
sana-jawad: Make sense. Testing here for minPartition to be empty is a useless check. I will remove it. Checking for error to be not nil is good enough. And since same logic applies to maxPartition, will do the same thing there too.

pkg/sloop/common/db_utilities_test.go
```diff
@@ -12,46 +12,29 @@ var commonPrefix = "/commonprefix/001546405200/"
...
+	assert.Equal(t, uint64(0), numOfDeletedKeys)
+	assert.Equal(t, uint64(0), numOfKeysToDelete)
 }
 
 func Test_Db_Utilities_DeleteKeysWithPrefix_DeleteSomeKeys(t *testing.T) {
 	db := helper_get_db(t)
 	helper_add_keys_to_db(t, db, helper_testKeys_with_common_prefix(commonPrefix))
 	helper_add_keys_to_db(t, db, helper_testKeys_with_common_prefix("randomStuff"+commonPrefix))
-	err, numOfDeletedKeys, numOfKeysToDelete := DeleteKeysWithPrefix([]byte(commonPrefix), db, 10)
+	err, numOfDeletedKeys, numOfKeysToDelete := DeleteKeysWithPrefix(commonPrefix, db, 10, 4)
```
mallow111, line 34: This test is almost the same as Test_Db_Utilities_DeleteKeysWithPrefix_DeleteAllKeys, maybe we can get count of remain keys to make sure we did not delete all keys.

pkg/sloop/common/partition_utilities.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// prints all the keys histogram. It can help debugging when needed.
+func PrintKeyHistogram(db badgerwrap.DB) {
+	partitionTableNameToKeyCountMap, totalKeyCount := GetPartitionsInfo(db)
+	glog.Infof("TotalkeyCount: %v", totalKeyCount)
+
+	for partitionID, partitionInfo := range partitionTableNameToKeyCountMap {
+		for tableName, keyCount := range partitionInfo.TableNameToKeyCountMap {
+			glog.Infof("TableName: %v, PartitionId: %v, keyCount: %v", tableName, partitionID, keyCount)
```
mallow111, line 40: This will print out a lot of keys by default, maybe we can add glog.V(some#)so we only print when needed.

pkg/sloop/common/partition_utilities.go
```diff
@@ -0,0 +1,87 @@
...
+	_ = db.View(func(txn badgerwrap.Txn) error {
+		iterOpt := badger.DefaultIteratorOptions
+		iterOpt.PrefetchValues = false
+		it := txn.NewIterator(iterOpt)
+		defer it.Close()
+		for it.Rewind(); it.Valid(); it.Next() {
+			item := it.Item()
+			sloopKey, err := GetSloopKey(item)
+			if err != nil {
+				glog.Infof("failed to parse information about key: %x", item.Key())
```
mallow111, line 73: when err != nil, should we continue to the next item?

pkg/sloop/storemanager/storemanager.go
sana-jawad: its already logged in DeleteKeysWithPrefix()

pkg/sloop/common/db_utilities.go
sana-jawad: I have kept string because its easier to view when debugging and change it to byte[] only when passing to the function which deletes/drops

pkg/sloop/common/partition_utilities.go
sana-jawad: yes in case there is just one bad key which got added somehow. I would continue with the execution of main logic.

pkg/sloop/common/db_utilities_test.go
sana-jawad: The challenge with delete keys is that its not on a locked db which can lead to a race condition. For example, the keys are getting added for the prefix that you are trying to delete. So in order to avoid it, we return as soon as the number of keys deleted is the desired number. Let me add a comment to add it more clear that in this test case the DB has 8 keys and we are only deleting 4.

pkg/sloop/storemanager/storemanager.go
mallow111: No, it  did not, I see return err, where is the log part?

pkg/sloop/common/db_utilities.go
mallow111: Can we change  keyPrefix []byte to  keyPrefix String ? From the code maintenance perspective,  we want to keep it consistent. When we call a library, we don't want to always think when to pass string and when to pass []byte, especially this is open sourced  project.

pkg/sloop/common/partition_utilities.go
mallow111: if the only key is a bad key, the following logic should not work, and if you have 10000 keys, and 90% are bad, by adding `continue` you dont need to do the following calculation. Also your totalKeyCount should be valid key count. partitionIDToPartitionInfoMap should only contains valid key paritionID.

mallow111:
left a few comments

pkg/sloop/common/partition_utilities.go
sana-jawad: added the continue

PR #119, opened 2020-05-11
Read me for memory consumption

sana-jawad:
Added guidelines in read me for memory consumption.
Also added metric for measuring dropPrefix latency

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ Apart from these flags some other values can be tweaked to fit in the memory con
 
 - Memory consumption max limit: 1GB
 ``` 
-               // 0.5<<20 (524288 bytes = 1 Mb)               
+               // 0.5<<20 (524288 bytes = 0.5 Mb)               
                "badger-max-table-size=524288",
                "badger-number-of-compactors=1",
                "badger-number-of-level-zero-tables=1",
--- a/README.md
+++ b/README.md
@@ -97,6 +97,41 @@ To download a backup of the database, navigate to http://localhost:8080/data/bac
 
 To restore from a backup, start `sloop` with the `-restore-database-file` flag set to the backup file downloaded in the previous step. When restoring, you may also wish to set the `-disable-kube-watch=true` flag to stop new writes from occurring and/or the `-context` flag to restore the database into a different context.
 
+## Memory Consumption
+
+Sloop's memory usage can be managed by tweaking several options:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -97,6 +97,42 @@ To download a backup of the database, navigate to http://localhost:8080/data/bac
...
+- `badger-use-lsm-only-options` If this flag is set to true, values would be collocated with the LSM tree, with value log largely acting as a write-ahead log only. Recommended value for memory constrained environments: false
+- `badger-keep-l0-in-memory` When this flag is set to true, Level 0 tables are kept in memory. This leads to better performance in writes as well as compactions. Recommended value for memory constrained environments: false
+- `badger-sync-writes` When SyncWrites is true all writes are synced to disk. Setting this to false would achieve better performance, but may cause data loss in case of crash. Recommended value for memory constrained environments: false
+- `badger-vlog-fileIO-mapping` TableLoadingMode indicates which file loading mode should be used for the LSM tree data files. Setting to true would not load the value in memory map. Recommended value for memory constrained environments: true
+
+Apart from these flags some other values can be tweaked to fit in the memory constraints. Following are some examples of setups.
+
+- Memory consumption max limit: 1GB
+``` 
+               // 0.5<<20 (524288 bytes = 1 Mb)
```
duke-harlan, line 112: this `(524288 bytes = 1 Mb)` seems wrong

README.md
```diff
@@ -97,6 +97,42 @@ To download a backup of the database, navigate to http://localhost:8080/data/bac
...
+``` 
+               // 0.5<<20 (524288 bytes = 1 Mb)               
+               "badger-max-table-size=524288",
+               "badger-number-of-compactors=1",
+               "badger-number-of-level-zero-tables=1",
+               "badger-number-of-zero-tables-stall=2",
+```
+- Memory consumption max limit: 2GB
+``` 
+               // 16<<20 (16777216 bytes = 16 Mb)
```
duke-harlan, line 120: is it evident somewhere else how 16Mb translates to 2GB?

README.md
sana-jawad: yes it should be 0.5 MB

README.md
sana-jawad: So the max table size helps in cleanup. If we have less data, the table size needs to be smaller. The size of DB affects the memory consumed, so if size is in check, the memory consumed is also in check. We arrived at these values after a number of test runs.

Issue #116, opened 2020-04-29
Sloop Releases Request


acaire:
Hi, any chance you could please use releases and subsequently immutable style versioning for docker tags? I'd like to be able to pin my Sloop deployments rather than using latest.

I'm happy to submit a PR for this but would need some direction in the implementation as it requires buy-in from your side.

sana-jawad:
Thanks @acaire we have added it in our backlog and will get to it soon.

mallow111:
https://github.com/salesforce/sloop/pull/121/files is merged, will close this issue.

chelomontilla:
Only latest tag can be found in https://hub.docker.com/r/sloopimage/sloop. Some more changes required?

PR #121, opened 2020-06-17
using goreleaser for semver builds on sloop. 

sridhav:
Integrated with travis CI and docker.

fixes issue #116

Test builds: https://travis-ci.org/github/sridhav/sloop/builds -- the tag builds will fail as i don't have permissions to push to `salesforce/sloop` docker image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,12 @@
-FROM golang:1.13 AS build
-
-WORKDIR /sloop
-COPY go.mod go.sum ./
-
-RUN go mod download
-
-COPY pkg ./pkg
-
-RUN curl -o /sloop/aws-iam-authenticator https://amazon-eks.s3-us-west-2.amazonaws.com/1.14.6/2019-08-22/bin/linux/amd64/aws-iam-authenticator \
+FROM alpine as build
+RUN apk add --no-cache curl
+RUN curl -o /aws-iam-authenticator https://amazon-eks.s3-us-west-2.amazonaws.com/1.14.6/2019-08-22/bin/linux/amd64/aws-iam-authenticator \
   && wait \
-  && chmod +x /sloop/aws-iam-authenticator
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags "-s" -installsuffix cgo -o sloop ./pkg/sloop
+  && chmod +x /aws-iam-authenticator
 
 FROM gcr.io/distroless/base
-COPY --from=build /sloop/sloop /sloop
-COPY --from=build /sloop/pkg/sloop/webfiles /pkg/sloop/webfiles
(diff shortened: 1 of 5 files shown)


sridhav:
FYI @nurland @mallow111 @sana-jawad

Dockerfile
```diff
@@ -1,20 +1,12 @@
...
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags "-s" -installsuffix cgo -o sloop ./pkg/sloop
+  && chmod +x /aws-iam-authenticator
 
 FROM gcr.io/distroless/base
-COPY --from=build /sloop/sloop /sloop
-COPY --from=build /sloop/pkg/sloop/webfiles /pkg/sloop/webfiles
-COPY --from=build /sloop/aws-iam-authenticator /aws-iam-authenticator
+COPY sloop /sloop
+COPY pkg/sloop/webfiles /pkg/sloop/webfiles
+COPY --from=build /aws-iam-authenticator /aws-iam-authenticator
```
sana-jawad, line 10: what will be the process of building image locally. Can you please update the READ ME as a part of this PR with the instructions.

sridhav: @sana-jawad Its similar to earlier, just requires running `make docker`, which should build binary and the image itself

Dockerfile
```diff
@@ -1,20 +1,12 @@
...
-COPY go.mod go.sum ./
-
-RUN go mod download
-
-COPY pkg ./pkg
-
-RUN curl -o /sloop/aws-iam-authenticator https://amazon-eks.s3-us-west-2.amazonaws.com/1.14.6/2019-08-22/bin/linux/amd64/aws-iam-authenticator \
+FROM alpine as build
+RUN apk add --no-cache curl
+RUN curl -o /aws-iam-authenticator https://amazon-eks.s3-us-west-2.amazonaws.com/1.14.6/2019-08-22/bin/linux/amd64/aws-iam-authenticator \
```
sana-jawad, line 3: why do we have this path change?

Makefile
```diff
@@ -35,3 +39,11 @@ protobuf:
 cover:
 	go test ./pkg/... -coverprofile=coverage.out
 	go tool cover -html=coverage.out
+
+release:
+	@if [ ! -z "$(GITHUB_TOKEN)" ];then \
```
sana-jawad, line 44: Will it work for anyone's GITHUB_TOKEN?

sridhav: @sana-jawad you need to have salesforce github token to push releases to salesforce repository. If we are unable to to get that token we skip artifact publish and just push the docker image.

Dockerfile
sridhav: @sana-jawad Initially in the docker file we were using two images one to build the binary and second is the actual sloop image. As the build process is entirely taken care by goreleaser we just moved to the root path directly. instead of getting build data from other image

PR #122, opened 2020-06-18
fixed scrolling issue on the sideleftnav

sridhav:
Left navigation is not scrollable, fixed by updating css


sridhav:
@sana-jawad @mallow111

PR #124, opened 2020-08-31
fix the deployment kind missing issue and add address binding feature

mallow111:
This PR is trying to fix 2 things:
1. Our current sloop does not show deployment, this PR is trying to fix this issue.
2. Make bind-address configurable so that when we add oauth2-proxy in front of sloop, we can pass ip address to sloop.


pkg/sloop/server/internal/config/config.go
```diff
@@ -33,6 +33,7 @@ type SloopConfig struct {
 	DisableKubeWatcher       bool          `json:"disableKubeWatch"`
 	KubeWatchResyncInterval  time.Duration `json:"kubeWatchResyncInterval"`
 	WebFilesPath             string        `json:"webfilesPath"`
+	BindAddress              string        `json:"bindIp"`
```
nurland, line 36: nit: can you please make sure struct field name and json property name are in sync?

mallow111: good call, will do

PR #128, opened 2020-09-23
enable truncate value log option if badger db offset is different from badger db size

sridhav:
FYI @nurland @sana-jawad @mallow111

### DIFF
--- a/pkg/sloop/store/untyped/store.go
+++ b/pkg/sloop/store/untyped/store.go
@@ -95,7 +95,7 @@ func OpenStore(factory badgerwrap.Factory, config *Config) (badgerwrap.DB, error
 		opts = opts.WithValueLogLoadingMode(options.FileIO)
 	}
 
-	opts.Truncate = config.BadgerVLogTruncate
+	opts = opts.WithTruncate(config.BadgerVLogTruncate)
 
 	opts = opts.WithSyncWrites(config.BadgerSyncWrites)
 
--- a/pkg/sloop/store/untyped/store.go
+++ b/pkg/sloop/store/untyped/store.go
@@ -9,7 +9,7 @@ package untyped
 
 import (
 	"fmt"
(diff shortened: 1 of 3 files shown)


pkg/sloop/store/untyped/store.go
```diff
@@ -94,6 +95,8 @@ func OpenStore(factory badgerwrap.Factory, config *Config) (badgerwrap.DB, error
 		opts = opts.WithValueLogLoadingMode(options.FileIO)
 	}
 
+	opts = opts.WithTruncate(config.BadgerVLogTruncate)
```
mallow111, line 98: one nit pick, I see other examples in this function give a if condition check before set opts, one example is: https://github.com/salesforce/sloop/blob/master/pkg/sloop/store/untyped/store.go#L93, maybe we should follow the same practice for code consistence.

sridhav: its a boolean value -> if clause is unnecessary op

sridhav: And regarding the consistency, line 100 below does the same, just assigns the variable as its a boolean

sana-jawad (commented):
Changes look good to me @sridhav. What testing has been done for this PR? Were you able to stimulate this error?

sridhav:
@sana-jawad I was unable to simulate this error in my local box, thats the reason i reached out in slack asking about integration testing? is there any integration test suite/stress test suite/chaos test suite used by sloop to simulate these kind of load based errors?

sana-jawad:
> @sana-jawad I was unable to simulate this error in my local box, thats the reason i reached out in slack asking about integration testing? is there any integration test suite/stress test suite/chaos test suite used by sloop to simulate these kind of load based errors?

Unfortunately we don't have a suite of integration tests but we do have saved data sets for this error in case it can help you in testing.

sana-jawad (approved):

PR #133, opened 2020-11-02
Helm values picked by Sloop

kritikapradhan:
problem -
even when the file is being provided, config values were being set to default and if a json file is inputted, sloop throws unmarshal errors

solution-
- added an if else clause to check if vals need to be taken from file or defaulted
- included unmarshalling for json files as inputs too
- added unit tests for the function


pkg/sloop/server/internal/config/config.go
```diff
@@ -172,16 +173,31 @@ func (c *SloopConfig) Validate() error {
...
-	yamlFile, err := ioutil.ReadFile(filename)
-	if err != nil {
-		panic(fmt.Sprintf("failed to read %v. %v", filename, err))
-	}
-	var config SloopConfig
-	err = yaml.Unmarshal(yamlFile, &config)
-	if err != nil {
-		panic(fmt.Sprintf("failed to unmarshal %v. %v", filename, err))
+	if strings.Contains(filename,".yaml") {
+		yamlFile, err := ioutil.ReadFile(filename)
```
sana-jawad, line 177: Can you please add unit tests for this case?

pkg/sloop/server/internal/config/config.go
```diff
@@ -172,16 +173,31 @@ func (c *SloopConfig) Validate() error {
...
+		if err != nil {
+			panic(fmt.Sprintf("failed to read %v. %v", filename, err))
+		}
+		var config SloopConfig
```
alekseymykhailov, line 181: maybe better to declare this before if statement? so you don't need to define it twice (line 192).

pkg/sloop/server/internal/config/config.go
```diff
@@ -172,16 +173,31 @@ func (c *SloopConfig) Validate() error {
...
+		err = yaml.Unmarshal(yamlFile, &config)
```
alekseymykhailov, line 182: it looks like this is the only difference between json and yaml (line 182 vs line 193),  you could put only Unmarshal inside if statement, you don't need to duplicate error handling then

pkg/sloop/server/internal/config/config.go
kritikapradhan: I'll have to include two files with extensions .yaml and .json to the folder. what i can do is use filepath.ext(file_url)  to explain what this does?

pkg/sloop/server/internal/config/config.go
kritikapradhan: made the changes

pkg/sloop/server/internal/config/config.go
kritikapradhan: made the changes

pkg/sloop/server/internal/config/config.go
sana-jawad: yes you would need to add mock files to test this code

kritikapradhan: done!

pkg/sloop/server/internal/config/config_test.go
```diff
@@ -0,0 +1,33 @@
...
+	"encoding/json"
+	"github.com/ghodss/yaml"
+)
+
+func Test_loadFromJSONFile(t *testing.T) {
+	var expectedconfig SloopConfig
+	configfilename:="testconfig.json"
+	configFile, err := ioutil.ReadFile(configfilename)
+	err = json.Unmarshal(configFile, &expectedconfig)
+
```
sana-jawad, line 16: Add an assert here to make sure there is no error

pkg/sloop/server/internal/config/config_test.go
```diff
@@ -0,0 +1,33 @@
...
+	out_config,err := loadFromFile(configfilename)
+
+	assert.Nil(t,err)
+	assert.Equal(t,out_config,&expectedconfig)
+}
+
```
sana-jawad, line 22: Also add test cases for error cases like file not available, errors in marshalling etc

pkg/sloop/server/internal/config/testconfig.json
```diff
@@ -0,0 +1,44 @@
+{
+	"ConfigFile": "",
+	"leftBarLinks": null,
+	"resourceLinks": null,
+	"disableKubeWatch": false,
+	"kubeWatchResyncInterval": 1800000000000,
+	"webfilesPath": "./pkg/sloop/webfiles",
```
sana-jawad, line 7: please add a folder for mock data and these files there

sana-jawad (changes requested):
I would like to see a few more tests. Thanks!

pkg/sloop/server/internal/config/config_test.go
kritikapradhan: hi i agree we should thoroughly have checks to see if the file being passed is correct but i don't think they are part of this work item. i will create another work item for it though.

sana-jawad: The ask is to just add assert.Nil(t,err) at line 16. I don't think we should create WI for it.

kritikapradhan: done

pkg/sloop/server/internal/config/config_test.go
kritikapradhan: done

pkg/sloop/server/internal/config/config_test.go
```diff
@@ -25,9 +26,25 @@ func Test_loadFromYAMLFile(t *testing.T) {
...
 	assert.Nil(t,err)
 	assert.Equal(t,out_config,&expectedconfig)
 }
+
+func Test_loadFromTxtFile(t *testing.T) {
+	configfilename:="testconfig.txt"
+	_,err := loadFromFile(configfilename)
+
+	assert.Nil(t,err)
+
```
sana-jawad, line 42: nit: extra line

pkg/sloop/server/internal/config/config_test.go
```diff
@@ -25,9 +26,25 @@ func Test_loadFromYAMLFile(t *testing.T) {
...
+}
+func Test_loadFromNoFile(t *testing.T) {
+	configfilename:="config.json"
+	_,err := loadFromFile(configfilename)
+
+	assert.Nil(t,err)
+
```
sana-jawad, line 49: same here

pkg/sloop/server/internal/config/config_test.go
```diff
@@ -0,0 +1,50 @@
...
```
nurland, line 49: nit: remove empty lines

pkg/sloop/server/internal/config/testconfig.json
nurland: +1 to comment above, @kritikapradhan it is a good practice to move all test/mock data to a separate folder, so you won't accidentally use it in production.

pkg/sloop/server/internal/config/config_test.go
```diff
@@ -0,0 +1,50 @@
...
+	assert.Nil(t,err)
+	assert.Equal(t,out_config,&expectedconfig)
+}
+
+func Test_loadFromTxtFile(t *testing.T) {
+	configfilename:="testconfig.txt"
+	_,err := loadFromFile(configfilename)
+
+	assert.Nil(t,err)
```
nurland, line 41: formatting -> assert.Nil(t, err)
please run `make` it should run `go fmt` and format it for you.

**btw should this be assert.Panics**? Seems like loadFromFile function panics on error from these checks:

if strings.Contains(filename, ".yaml") {
		err = yaml.Unmarshal(configFile, &config)
	} else if strings.Contains(filename, ".json"){
		err = json.Unmarshal(configFile, &config)

pkg/sloop/server/internal/config/config_test.go
```diff
@@ -0,0 +1,50 @@
...
+	err = yaml.Unmarshal(configFile, &expectedconfig)
+	assert.Nil(t,err)
+
+	out_config,err := loadFromFile(configfilename)
+
+	assert.Nil(t,err)
+	assert.Equal(t,out_config,&expectedconfig)
+}
+
+func Test_loadFromTxtFile(t *testing.T) {
```
nurland, line 37: please use the following naming for test functions: `Test_<action>_<expectedResult>`
for example: `Test_loadFromTextFile_Failure`, `Test_loadFromJson_Success`.

nurland (changes requested):

pkg/sloop/server/internal/config/config_test.go
mallow111: rather than creating this file, should we just put the json/yaml file content as a constant and calling it in the unit test, such as 
```
test= `
{
	"ConfigFile": "",
	"leftBarLinks": null,
	"resourceLinks": null,
	"disableKubeWatch": false,
	"kubeWatchResyncInterval": 1800000000000,
	"webfilesPath": "./pkg/sloop/webfiles",
	"bindAddress": "",
	"port": 8080,
	"storeRoot": "./data",
	"maxLookBack": 12
...
}
`
```
This way we will have more testing case and dont need to create any of the files, also i would recommend we put the unit test in a separate PR or work item, so we get this bug rolled in to unblock customers, also I think our ic needs sometime to learn how to use mock probably, that might take her some time to learn and write code

pkg/sloop/server/internal/config/config.go
```diff
@@ -171,17 +172,27 @@ func (c *SloopConfig) Validate() error {
 	return nil
 }
 
-func loadFromFile(filename string) *SloopConfig {
-	yamlFile, err := ioutil.ReadFile(filename)
+func loadFromFile(filename string) (*SloopConfig, error) {
```
mallow111, line 175: if we are going to use panic, this error seems only return nil, we can just return *SloopConfig

sana-jawad (dismissed):

pkg/sloop/server/internal/config/config_test.go
sana-jawad: extra lines are still there

pkg/sloop/server/internal/config/config.go
nurland: +1 to above, this function never returns error, let's update the signature and return *SloopConfig only.

nurland (dismissed):

nurland (dismissed):

pkg/sloop/server/internal/config/config.go
```diff
@@ -172,16 +173,26 @@ func (c *SloopConfig) Validate() error {
...
+	var config SloopConfig
+
+	configFile, err := ioutil.ReadFile(filename)
 	if err != nil {
 		panic(fmt.Sprintf("failed to read %v. %v", filename, err))
 	}
-	var config SloopConfig
-	err = yaml.Unmarshal(yamlFile, &config)
+
+	if strings.Contains(filename, ".yaml") {
```
duke-harlan, line 183: consider the paradigm of trying the `yaml.Unmarshal`, if it works then use it. If it fails, then try `json.Unmarshal`. If that fails, then report the error. This avoids the requirement of file extensions.

pkg/sloop/server/internal/testFiles/testConfig.txt
```diff
@@ -0,0 +1,44 @@
+{
+	"ConfigFile": "",
+	"leftBarLinks": null,
+	"resourceLinks": null,
+	"disableKubeWatch": false,
+	"kubeWatchResyncInterval": 1800000000000,
```
duke-harlan, line 6: consider using something like `180m` instead of some large number of nanoseconds
repeat for other `time.Duration` values

pkg/sloop/server/internal/testFiles/testConfig.txt
```diff
@@ -0,0 +1,44 @@
+{
```
duke-harlan, line 1: unless there is a separate test that covers this - make sure that most/all of these are not the same as default values - want to ensure not accidentally getting the wrong config

pkg/sloop/server/internal/config/config_test.go
```diff
@@ -0,0 +1,42 @@
...
+	configFile, err := ioutil.ReadFile(configfilename)
+	err = yaml.Unmarshal(configFile, &expectedconfig)
+	assert.Nil(t, err)
+
+	out_config := loadFromFile(configfilename)
+	assert.Equal(t, out_config, &expectedconfig)
+}
+
+func Test_loadFromTxtFile_shouldPanic(t *testing.T) {
+	configfilename, _ := filepath.Abs("../testFiles/testconfig.txt")
```
duke-harlan, line 35: why should a file with `.txt` extension be unusable

duke-harlan (changes requested):
I'm suggesting a paradigm change - switch from extension based behavior to file contents behavior

duke-harlan (approved):

nurland (approved):

PR #135, opened 2020-12-30
Rafabios feature/ingress-and-svc-type

rafabios:
Added ingress templating suppot: is possible to use nginx, kong, traefik and so on as a ingress controller;

Added service type, so users can choose between ClusterIP, NodePort and LoadBalancer;

### DIFF
--- a/helm/sloop/templates/ingress.yaml
+++ b/helm/sloop/templates/ingress.yaml
@@ -1,0 +1,29 @@
+{{- if .Values.ingress.enabled }}
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  annotations:
+  {{- range $key, $value := .Values.ingress.annotations }}
+    {{ $key }}: {{ $value | quote }}
+  {{- end }}
+  labels:
+    app.kubernetes.io/name: {{ .Values.name }}
+    helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    app.kubernetes.io/instance: "{{ .Release.Name }}"
+    app.kubernetes.io/managed-by: "{{ .Release.Service }}"
+  name: {{ .Values.name }}
+  #namespace: {{ .Values.namespace }}
+spec:
+  rules:
+  - host: {{ .Values.ingress.host }}
+    http:
+      paths:
+      - backend:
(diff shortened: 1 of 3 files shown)


helm/sloop/templates/ingress.yaml
```diff
@@ -0,0 +1,29 @@
...
+  {{- range $key, $value := .Values.ingress.annotations }}
+    {{ $key }}: {{ $value | quote }}
+  {{- end }}
+  labels:
+    app.kubernetes.io/name: {{ .Values.name }}
+    helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    app.kubernetes.io/instance: "{{ .Release.Name }}"
+    app.kubernetes.io/managed-by: "{{ .Release.Service }}"
+  name: {{ .Values.name }}
+  #namespace: {{ .Values.namespace }}
```
mallow111, line 15: either remove this comment line or add namespace variables in Values file

helm/sloop/templates/ingress.yaml
```diff
@@ -0,0 +1,29 @@
...
+  rules:
+  - host: {{ .Values.ingress.host }}
+    http:
+      paths:
+      - backend:
+          serviceName: {{ .Values.name }}
+          servicePort: {{ .Values.service.port }}
+        path: /
+{{- end -}}
+  {{- if .Values.ingress.tls }}
```
mallow111, line 26: just curious, ingress.tls is a {}, how does this work in if condition?

helm/sloop/templates/ingress.yaml
rafabios: Done! removed namespace. Sorry about that!

helm/sloop/templates/ingress.yaml
rafabios: if you remove {} it should accept kubernetes ingress format, like this:

```
 tls:
  - secretName: sloop-tls
    hosts:
     - sloop.example.com
```

Also a examplo from kubernetes-dashboard:
https://github.com/helm/charts/blob/7e45e678e39b88590fe877f159516f85f3fd3f38/stable/kubernetes-dashboard/values.yaml#L149

sana-jawad (approved):

PR #138, opened 2021-02-04
Make relative paths work with Sloop

sana-jawad:
Problem: Currently paths in Sloop are absolute hence http://sloop.com would work one cluster.
But if we want to support multiple Sloop instances using same address like http://sloop.com/clusterA and http://sloop.com/clusterB, this cannot work with current absolute paths. 
For example if we hit http://sloop.com/clusterA and enter a search query. It takes to  http://sloop.com/?query=EventHeatMap and thus the context of clusterA gets lost.
This PR addresses this issue.


pkg/sloop/webserver/webserver.go
```diff
@@ -167,30 +171,45 @@ func healthHandler() http.HandlerFunc {
...
-func Run(config WebConfig, tables typed.Tables) error {
-	webFiles = config.WebFilesPath
-	server := &Server{}
-	server.mux = http.NewServeMux()
-	server.mux.HandleFunc("/webfiles/", webFileHandler)
-	server.mux.HandleFunc("/data/backup", backupHandler(tables.Db(), config.CurrentContext))
-	server.mux.HandleFunc("/data", queryHandler(tables, config.MaxLookback))
-	server.mux.HandleFunc("/resource", resourceHandler(config.ResourceLinks))
+// Handler for redirecting / to /currentContext to ensure backward compatibility
+func redirectHandler(currentContext string) http.HandlerFunc {
```
mallow111, line 175: please add unit test

pkg/sloop/webserver/webserver.go
```diff
@@ -167,30 +171,45 @@ func healthHandler() http.HandlerFunc {
...
+	return func(writer http.ResponseWriter, request *http.Request) {
+		redirectURL := "/"+currentContext
+		http.Redirect(writer, request, redirectURL, http.StatusTemporaryRedirect)
+	}
+}
+
+// Registers paths for mux router
+func registerPaths(router *mux.Router, config WebConfig, tables typed.Tables) {
+	router.PathPrefix("/webfiles/").HandlerFunc(webFileHandler(config.CurrentContext))
```
mallow111, line 184: any reason this line use router.PathPrefix, while the rest use router.HandleFunc, can we all use router.HandleFunc

pkg/sloop/webserver/webserver.go
```diff
@@ -90,25 +91,28 @@ func logWebError(err error, note string, r *http.Request, w http.ResponseWriter)
...
-		return
-	}
-	w.Header().Set("content-type", mime.TypeByExtension(filepath.Ext(fullPath)))
-	_, err = w.Write(data)
-	if err != nil {
-		logWebError(err, "Error writing web file: "+fixedUrl, r, w)
-		return
+func webFileHandler(currentContext string) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		webFilesPathToTrim := "/" + currentContext + "/webfiles"
```
mallow111, line 96: what if currentContext is "" ?

pkg/sloop/webserver/webserver.go
sana-jawad: Yes that is because the full path of static files will be a whole path of it's own for example /webfiles/index.html. In this case all requests targeting static files should target webfiles handler. This works by using the prefix router method.
For details https://github.com/gorilla/mux#static-files

pkg/sloop/webserver/webserver.go
sana-jawad: Its the context for which Sloop is deployed. It would always be there.

pkg/sloop/webserver/webserver.go
sana-jawad: Sure added

pkg/sloop/webserver/webserver.go
sana-jawad: added unit test

pkg/sloop/webserver/webfiles/resource.html
```diff
@@ -8,9 +8,9 @@
 <html lang="en">
 <head>
     <meta charset="UTF-8">
-    <link rel='shortcut icon' type='image/x-icon' href='/webfiles/favicon.ico'/>
-    <link rel="stylesheet" type="text/css" href="/webfiles/sloop.css">
-    <link rel="stylesheet" type="text/css" href="/webfiles/resource.css">
+    <link rel='shortcut icon' type='image/x-icon' href='webfiles/favicon.ico'/>
```
duke-harlan, line 11: it looks like all the other `.html` files are getting a script inserted - not this one?

pkg/sloop/webserver/webserver.go
```diff
@@ -90,25 +91,28 @@ func logWebError(err error, note string, r *http.Request, w http.ResponseWriter)
...
-		return
-	}
-	w.Header().Set("content-type", mime.TypeByExtension(filepath.Ext(fullPath)))
-	_, err = w.Write(data)
-	if err != nil {
-		logWebError(err, "Error writing web file: "+fixedUrl, r, w)
-		return
+func WebFileHandler(currentContext string) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		webFilesPathToTrim := "/" + currentContext + "/webfiles"
```
duke-harlan, line 96: as suggested above - does `path.join()` work here

duke-harlan (approved):
looks ok - a couple of minor comments

pkg/sloop/webserver/webfiles/resource.html
sana-jawad: good observation :) yes resource.html is within main page. As the base href was set for the main page...its not needed by resource.html

pkg/sloop/webserver/webfiles/filter.js
```diff
@@ -89,11 +99,12 @@ function setFiltersAndReturnQueryUrl(defaultLookback, defaultKind, defaultNamesp
 
     namematch = setText("namematch", "filternamematch", "")
 
-    query =           populateDropdownFromQuery("query",     "filterquery",     "EventHeatMap",  "/data?query=Queries&lookback="+lookback);
-    ns =              populateDropdownFromQuery("namespace", "filternamespace", defaultNamespace, "/data?query=Namespaces&lookback="+lookback);
-    kind =            populateDropdownFromQuery("kind",      "filterkind",      defaultKind,      "/data?query=Kinds&lookback="+lookback);
+    windowLocation = getWindowLocationPathName()
```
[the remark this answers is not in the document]
sana-jawad: Yes got rid of the method

mallow111 (dismissed):

pkg/sloop/webserver/resourcehandler.go
```diff
@@ -83,10 +84,10 @@ func resourceHandler(resLinks []ResourceLinkTemplate) http.HandlerFunc {
 		queryEnd := d.ClickTime.Add(d.PlusMinusTime).Unix()
 
 		dataParams := fmt.Sprintf("?query=%v&namespace=%v&start_time=%v&end_time=%v&kind=%v&name=%v", "GetEventData", d.Namespace, queryStart, queryEnd, d.Kind, d.Name)
-		d.EventsUrl = "/data" + dataParams
+		d.EventsUrl = path.Join("/", currentContext, "data" + dataParams)
```
duke-harlan, line 87: Are `dataParams` actually part of the `path`? If everything past the `?` is considered params for the call, then we should move the `+ dataParams` outside of the call to `path.Join`. If you think of the entire URL as the path, then it is fine where it is.

duke-harlan (approved):
1 nit comment

pkg/sloop/webserver/resourcehandler.go
sana-jawad: yeah I would like to keep it here for clarity purposes

PR #139, opened 2021-02-05
Updated readme, added how to update bindata.go

kritikapradhan:


sana-jawad (approved):
Thanks for updating ReadMe. It would be great if we can make it a part of make run, so that developers don't need this step going forward @kritikapradhan

PR #140, opened 2021-02-11
@W-8881273 - Clean up webfiles based on learnings in training

kritikapradhan:


pkg/sloop/webserver/webfiles.go
```diff
@@ -6,33 +6,29 @@ import (
...
 	if err == nil {
 		return data, err
-	}
-	files := AssetNames()
-	//if file exists in binary form
-	if common.Contains(files, filepath) {
-		return Asset(filepath)
 	} else {
-		return nil, err
+		//file exists in binary
```
duke-harlan, line 25: `else {` above is not needed and can reduce indentation of following code

pkg/sloop/webserver/webfiles.go
```diff
@@ -6,33 +6,29 @@ import (
...
+		binFileList := AssetNames()
+		if common.Contains(binFileList, common.GetFilePath(prefix, fileName)) {
```
duke-harlan, line 27: consider a local variable `filepath := common.GetFilePath(prefix, fileName)` so they are combined only once

pkg/sloop/webserver/webfiles.go
```diff
@@ -41,14 +37,11 @@ func readWebfile(filepath string, fs *afero.Afero) ([]byte, error) {
...
-	data, err := readWebfile((path.Join(prefix, templateName)), &fs)
-	if err != nil {
-		return nil, err
-	}
-	newTemplate := template.New(templateName)
-	newTemplate, err = newTemplate.Parse(string(data))
-	if err != nil {
-		return nil, err
+	data, err := readWebfile(templateName, &fs)
+	if err == nil {
```
duke-harlan, line 41: the usual pattern of `if err != nil {` should be followed unless strong reason not too. It leaves the main line code at the least indentation

pkg/sloop/webserver/webfiles.go
```diff
@@ -41,14 +37,11 @@ func readWebfile(filepath string, fs *afero.Afero) ([]byte, error) {
...
+		newTemplate := template.New(templateName)
+		newTemplate, err = newTemplate.Parse(string(data))
```
duke-harlan, line 43: this `err` is being ignored

pkg/sloop/webserver/webserver.go
```diff
@@ -99,8 +100,8 @@ func webFileHandler(currentContext string) http.HandlerFunc {
 			logWebError(nil, "Not allowed", r, w)
 			return
 		}
-		fullPath := path.Join(prefix, fixedUrl)
-		data, err := readWebfile(fullPath, &afero.Afero{afero.NewOsFs()})
+		fullPath := common.GetFilePath(prefix, fixedUrl)
```
duke-harlan, line 103: if `fullPath` still needed?

duke-harlan (commented):
handful of comments in line

pkg/sloop/webserver/webserver.go
kritikapradhan: yes still needed, i moved it right above the line it's being used.

pkg/sloop/webserver/webfiles.go
```diff
@@ -6,33 +6,29 @@ import (
...
+// sample input : fileName="index.html"
+func readWebfile(fileName string, fs *afero.Afero) ([]byte, error) {
+	//file exists as a physical file
+	data, err := fs.ReadFile(common.GetFilePath(webFilesPath, fileName))
+	if err != nil {
+		//file exists in binary
+		binFileList := AssetNames()
+		binFileName := common.GetFilePath(prefix, fileName)
+		if common.Contains(binFileList, binFileName) {
+			return Asset(binFileName)
```
duke-harlan, line 27: could all of this block of code be condenced to:
```
    return Asset(common.GetFilePath(prefix, fileName))
```
this is after seeing the `Asset()` is checking filenames, and assuming that the error message works (not confusing & contains the asset name)

duke-harlan (approved):
looks ok - 1 suggestion
- do we need to check the existence of the binary form before calling `Asset()`

Issue #62, opened 2019-11-14
Sloop will not pick up CRD types created after the start of sloop, gives errors when CRD types are removed after startup


thomashargrove:
Sloop scans the list of CRDs at startup, then sets up watches for each CRD type.  If CRD types are registered later Sloop does not recognize them.  If CRD types are deleted from the server Sloop will return errors like:

E1114 09:59:37.508684   10297 reflector.go:153] pkg/mod/k8s.io/client-go@v0.0.0-20191003035859-a746c2f219b7/tools/cache/reflector.go:105: Failed to list *unstructured.Unstructured: the server could not find the requested resource

PR #142, opened 2021-02-12
Refactor CRD watcher to allow refreshing (drop old, create new)

duke-harlan:
https://github.com/salesforce/sloop/issues/62

P: kubewatcher creates CRD informers only at startup
S: periodically check CRD list - start new Informers as needed & stop unwanted informers

kubeWatcherImpl keeps a map entry for each active CRD informer
each active CRD informer has its own stop channel
a timer is started that will refresh the CRD Informers as needed
2 new metrics to track started & running - should be the same!

Extras:
update some of the static watchers to a more recent schema hierarchy point
shuffle some stuff to make the code more testable
leverage k8s fakes for some unit tests
get CRD list will handle two schema versions
oh, and implement some 'unit' tests

### DIFF
--- a/pkg/sloop/ingress/kubewatcher.go
+++ b/pkg/sloop/ingress/kubewatcher.go
@@ -91,7 +91,7 @@ func NewKubeWatcherSource(kubeClient kubernetes.Interface, outChan chan typed.Ku
 		}
 
 		kw.refreshCrd = time.NewTicker(crdRefreshInterval)
-		go kw.refershCrdInformers(masterURL, kubeContext)
+		go kw.refreshCrdInformers(masterURL, kubeContext)
 	}
 
 	return kw, nil
@@ -332,7 +332,7 @@ func (i *kubeWatcherImpl) getResourceAsJsonString(kind string, obj interface{}) 
 	return string(bytes), nil
 }
 
-func (i *kubeWatcherImpl) refershCrdInformers(masterURL string, kubeContext string) {
+func (i *kubeWatcherImpl) refreshCrdInformers(masterURL string, kubeContext string) {
 	for _ = range i.refreshCrd.C {
 		glog.Infof("Starting to refresh CRD informers")
 		err := i.startCustomInformers(masterURL, kubeContext)
--- a/pkg/sloop/ingress/kubewatcher.go
+++ b/pkg/sloop/ingress/kubewatcher.go
@@ -100,10 +100,14 @@ func NewKubeWatcherSource(kubeClient kubernetes.Interface, outChan chan typed.Ku
 func (i *kubeWatcherImpl) startWellKnownInformers(kubeclient kubernetes.Interface) {
 	i.informerFactory = informers.NewSharedInformerFactory(kubeclient, i.resync)
 
+	i.informerFactory.Apps().V1().DaemonSets().Informer().AddEventHandler(i.getEventHandlerForResource("DaemonSet"))
 	i.informerFactory.Apps().V1().Deployments().Informer().AddEventHandler(i.getEventHandlerForResource("Deployment"))
+	i.informerFactory.Apps().V1().ReplicaSets().Informer().AddEventHandler(i.getEventHandlerForResource("ReplicaSet"))
 	i.informerFactory.Apps().V1().StatefulSets().Informer().AddEventHandler(i.getEventHandlerForResource("StatefulSet"))
+
 	i.informerFactory.Core().V1().ConfigMaps().Informer().AddEventHandler(i.getEventHandlerForResource("ConfigMap"))
 	i.informerFactory.Core().V1().Endpoints().Informer().AddEventHandler(i.getEventHandlerForResource("Endpoint"))
+	i.informerFactory.Core().V1().Events().Informer().AddEventHandler(i.getEventHandlerForResource("Event"))
 	i.informerFactory.Core().V1().Namespaces().Informer().AddEventHandler(i.getEventHandlerForResource("Namespace"))
 	i.informerFactory.Core().V1().Nodes().Informer().AddEventHandler(i.getEventHandlerForResource("Node"))
 	i.informerFactory.Core().V1().PersistentVolumeClaims().Informer().AddEventHandler(i.getEventHandlerForResource("PersistentVolumeClaim"))
@@ -111,14 +115,14 @@ func (i *kubeWatcherImpl) startWellKnownInformers(kubeclient kubernetes.Interfac
 	i.informerFactory.Core().V1().Pods().Informer().AddEventHandler(i.getEventHandlerForResource("Pod"))
 	i.informerFactory.Core().V1().Services().Informer().AddEventHandler(i.getEventHandlerForResource("Service"))
 	i.informerFactory.Core().V1().ReplicationControllers().Informer().AddEventHandler(i.getEventHandlerForResource("ReplicationController"))
-	i.informerFactory.Extensions().V1beta1().DaemonSets().Informer().AddEventHandler(i.getEventHandlerForResource("DaemonSet"))
-	i.informerFactory.Extensions().V1beta1().ReplicaSets().Informer().AddEventHandler(i.getEventHandlerForResource("ReplicaSet"))
+
 	i.informerFactory.Storage().V1().StorageClasses().Informer().AddEventHandler(i.getEventHandlerForResource("StorageClass"))
-	i.informerFactory.Core().V1().Events().Informer().AddEventHandler(i.getEventHandlerForResource("Event"))
+
 	i.informerFactory.Start(i.stopChan)
 }
 
 func (i *kubeWatcherImpl) startCustomInformers(masterURL string, kubeContext string) error {
+
 	clientCfg := getConfig(masterURL, kubeContext)
 	kubeCfg, err := clientCfg.ClientConfig()
 	if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/sloop/ingress/kubewatcher.go
```diff
@@ -81,77 +100,149 @@ func NewKubeWatcherSource(kubeClient kubernetes.Interface, outChan chan typed.Ku
...
+		close(v.stopChan)
+	}
+}
+
+func getCrdList(crdClient clientset.Interface) ([]crdGroupVersionResourceKind, error) {
+	crdList, err := crdClient.ApiextensionsV1().CustomResourceDefinitions().List(metav1.ListOptions{})
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to instantiate client for querying CRDs")
+		glog.Errorf("Failed to get CRD list from ApiextensionsV1, falling back to ApiextensionsV1beta1")
+		return getCrdListV1beta1(crdClient)
```
kritikapradhan, line 218: What's the thought behind keeping 2 functionalities to get crd list ? safety net ?

pkg/sloop/ingress/kubewatcher.go
```diff
@@ -241,15 +332,30 @@ func (i *kubeWatcherImpl) getResourceAsJsonString(kind string, obj interface{})
 	return string(bytes), nil
 }
 
+func (i *kubeWatcherImpl) refershCrdInformers(masterURL string, kubeContext string) {
```
kritikapradhan, line 335: should it be 'refreshCRDInformers' ?

pkg/sloop/ingress/kubewatcher.go
```diff
@@ -73,6 +89,9 @@ func NewKubeWatcherSource(kubeClient kubernetes.Interface, outChan chan typed.Ku
 		if err != nil {
 			return nil, err
 		}
+
+		kw.refreshCrd = time.NewTicker(crdRefreshInterval)
+		go kw.refershCrdInformers(masterURL, kubeContext)
```
kritikapradhan, line 94: here as well 'refreshCRDInformers' ?

pkg/sloop/ingress/kubewatcher.go
duke-harlan: the schema changed - `getCrdListV1beta1` used the previous schema, this one is current.
If the environment that this code runs in is old, this method will fail and the other one should work

pkg/sloop/ingress/kubewatcher.go
duke-harlan: though technically correct (should capitalize the acronym), the acronym feels like (used as) a word and the camelCase seems appropriate.
Also the number of places that would need to change would make this PR harder to read.

pkg/sloop/ingress/kubewatcher.go
```diff
@@ -39,32 +40,47 @@ type KubeWatcher interface {
...
+	outchan        chan typed.KubeWatchResult
+	resync         time.Duration
+	protection     *sync.Mutex
+	stopped        bool
+	refreshCrd     *time.Ticker
+	currentContext string
+}
+
 var (
+	newCrdClient = func(kubeCfg *rest.Config) (clientset.Interface, error) { return clientset.NewForConfig(kubeCfg) }
```
mallow111, line 71: can we put this somewhere common so other pkg can also use? I feel this function will be used a lot in sloop

pkg/sloop/ingress/kubewatcher.go
```diff
@@ -81,77 +100,149 @@ func NewKubeWatcherSource(kubeClient kubernetes.Interface, outChan chan typed.Ku
...
 	if err != nil {
 		return errors.Wrap(err, "failed to read config while starting custom informers")
 	}
 
-	crdList, err := getCrdList(kubeCfg)
+	crdClient, err := newCrdClient(kubeCfg)
+	if err != nil {
+		return errors.Wrap(err, "failed to instantiate client for querying CRDs")
+	}
+	crdList, err := getCrdList(crdClient)
```
mallow111, line 136: do we want to put the err in a errors wrap and describe it failed because of get crd list?

pkg/sloop/ingress/kubewatcher.go
```diff
@@ -81,77 +100,149 @@ func NewKubeWatcherSource(kubeClient kubernetes.Interface, outChan chan typed.Ku
...
 	}
 
+	glog.Infof("Found %d CRD definitions", len(crdList))
 	dynamicClient, err := dynamic.NewForConfig(kubeCfg)
 	if err != nil {
 		return errors.Wrap(err, "failed to instantiate client for custom informers")
 	}
-
-	f := dynamicinformer.NewFilteredDynamicSharedInformerFactory(dynamicClient, i.resync, "", nil)
+	existing := i.pullCrdInformers()
```
mallow111, line 146: i would rename it as crdInformers

pkg/sloop/ingress/kubewatcher.go
```diff
@@ -81,77 +100,149 @@ func NewKubeWatcherSource(kubeClient kubernetes.Interface, outChan chan typed.Ku
...
+		glog.V(2).Infof("Stopping CRD informer for: %s (%v)", v.crd.kind, gvr)
+		close(v.stopChan)
+	}
+}
+
+func getCrdList(crdClient clientset.Interface) ([]crdGroupVersionResourceKind, error) {
+	crdList, err := crdClient.ApiextensionsV1().CustomResourceDefinitions().List(metav1.ListOptions{})
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to instantiate client for querying CRDs")
+		glog.Errorf("Failed to get CRD list from ApiextensionsV1, falling back to ApiextensionsV1beta1")
```
mallow111, line 217: do we want to include the err in the log

pkg/sloop/ingress/kubewatcher_test.go
```diff
@@ -0,0 +1,237 @@
...
+//      returns: clientset.Interface & error (always nil)
+func newTestCrdClient(reaction func(_ k8sTesting.Action) (bool, runtime.Object, error)) func(_ *rest.Config) (clientset.Interface, error) {
+	return func(_ *rest.Config) (clientset.Interface, error) {
+		crdClient := &clientsetFake.Clientset{}
+		crdClient.AddReactor("list", "*", reaction)
+		return crdClient, nil
+	}
+}
+
+func Test_bigPicture(t *testing.T) {
```
mallow111, line 66: what bigPicture mean here?

pkg/sloop/ingress/kubewatcher_test.go
```diff
@@ -0,0 +1,237 @@
...
+func reactionError(_ k8sTesting.Action) (bool, runtime.Object, error) {
+	return true, nil, fmt.Errorf("failed")
+}
+
+// newTestCrdClient - provides a function pointer to create a fake clientset
+//   takes: k8sTesting.Action function pointer - adds the reaction to the fake clientset
+//   returns a function pointer
+//      takes: restConfig
+//      returns: clientset.Interface & error (always nil)
+func newTestCrdClient(reaction func(_ k8sTesting.Action) (bool, runtime.Object, error)) func(_ *rest.Config) (clientset.Interface, error) {
```
mallow111, line 58: i feel this can be reused, do we want to put it somewhere common so other test can call?

pkg/sloop/ingress/kubewatcher.go
duke-harlan: if you're referring to the variable - `newCrdClient` this is a hook to allow UT to replace real client with mock/fake clients - sharing across packages can be a problem
if you're referring to the implementation - this is a very lightweight wrapper around the call to `NewForConfig` - this works better as an example instead of shared code

pkg/sloop/ingress/kubewatcher.go
duke-harlan: I wanted to express that there is more happening then just accessing the collection. The implementation of `pullCrdInformers` provides the collection of CRD informers, but leaves the in-object collection empty. This preps for the next step of refilling the collection with active/new CRD informers

pkg/sloop/ingress/kubewatcher.go
duke-harlan: as mentioned elsewhere, CRD is being used as a 'word' instead of an acronym

pkg/sloop/ingress/kubewatcher_test.go
duke-harlan: I prefer not to make something public/shared until there is at least another consumer. Early publishing of potential code clutters the global/shared namespace. The cost of making something shared later is not a significant barrier.

pkg/sloop/ingress/kubewatcher_test.go
duke-harlan: starts up the watcher, runs a k8s event through the system, then stop/cleanup.
This more of a test harness rather then a UT, but it validates that high level is successful.

duke-harlan (commented):
@mallow111 thanks for the comments. I responded to most. I'll make a couple of changes based on the others. Feel free to push back if you feel strongly about any of these issues.

pkg/sloop/ingress/kubewatcher.go
duke-harlan: done

pkg/sloop/ingress/kubewatcher.go
duke-harlan: done

pkg/sloop/ingress/kubewatcher.go
kritikapradhan: the spelling of refresh...

pkg/sloop/ingress/kubewatcher_test.go
mallow111: thanks for explanation, do you mind to put these as a comments in front of the function so people who have the similar question will be clear.

mallow111 (approved):

PR #143, opened 2021-02-16
Sloop: update steps to contribute page

mengyaoyang11:
Update following sections for Contribute page:
build
steps to contribute 
pull request check list

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,14 +1,63 @@
 ## Build
 
-Sloop uses GitHub to manager reviews of pull requests
+Sloop uses GitHub to manage reviews of pull requests
+
+[Singing up for a GitHub account](https://docs.github.com/en/github/getting-started-with-github/signing-up-for-a-new-github-account) if you don't have one
+
+
 
 ## Steps to Contribute
 
-ADD
+1.[Pick an issue you are interested in working on](https://github.com/salesforce/sloop/issues).
+It’s a good idea to comment on an Issue that you want to help with and ask some questions before doing any major coding work.
+
+2.Fork the repository: Look for the “Fork” button in the top right corner of [salesforce/sloop repo](https://github.com/salesforce/sloop)
+
+3.Clone the fork: Click on the green “clone or download” button,you can choose “clone with HTTPS.” Or, you can use SSH, which is compatible with 2-factor authentication, but just [some more work to set up](https://docs.github.com/en/github/authenticating-to-github/connecting-to-github-with-ssh).
+Go to the folder where you want to save the repository, type `git clone that-https-url-you-copied`
+
+4: Add a remote pointing to the original repository so you can sync changes: `git remote add upstream git@github.com:salesforce/sloop.git`
+
+
+5.Make a git branch: If you type git status you will see the branch name that you’re on, called master. In most projects, master is a special place where the most stable, reviewed, up-to-date code is. So, you’ll need to make your own branch: 
+
+`git branch a-descriptive-name`
+
+Then switch to that branch:
+`git checkout name-for-your-branch`
+
+6.Make some commits:
+
+When you have some code that you want to keep, you should save it in git by creating a commit. Here’s how:
+
+`git status` shows you the files you changed
+
+`git add path-to-your-file` allows you to pre-select the files you want to save.
+
+`git status` again to make sure you added the files you want to keep
+
+`git commit -m "some message here #123"` groups your changes together into a commit. The message should be short, describe the work that you did, and include the issue number that you are working on.
+
+`git push origin name-for-your-branch` to save your work online.
+
+7.Open a Pull Request: To create one, go to your fork of the project, click on the Pull Requests tab, and click the big green “New Pull Request” button. After you choose your branch,
+   Click the green “Create Pull Request” button. It will be super helpful if you can write a sentence or two summarizing the work you did and include a link to the Issue you were working on
+
+8.Expect changes and be patient: Next up, a maintainer or contributor will review your code. Once your work is approved, it will be merged in!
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,14 +1,63 @@
...
+`git commit -m "some message here #123"` groups your changes together into a commit. The message should be short, describe the work that you did, and include the issue number that you are working on.
+
+`git push origin name-for-your-branch` to save your work online.
+
+7.Open a Pull Request: To create one, go to your fork of the project, click on the Pull Requests tab, and click the big green “New Pull Request” button. After you choose your branch,
+   Click the green “Create Pull Request” button. It will be super helpful if you can write a sentence or two summarizing the work you did and include a link to the Issue you were working on
+
+8.Expect changes and be patient: Next up, a maintainer or contributor will review your code. Once your work is approved, it will be merged in!
+Congratulations and thank you for giving back to the open source community :) This whole thing only works because of people like you!
+
```
mallow111, line 48: I might remove  ```This whole thing only works because of people like you!```

CONTRIBUTING.md
```diff
@@ -1,14 +1,63 @@
...
+`git add path-to-your-file` allows you to pre-select the files you want to save.
+
+`git status` again to make sure you added the files you want to keep
+
+`git commit -m "some message here #123"` groups your changes together into a commit. The message should be short, describe the work that you did, and include the issue number that you are working on.
+
+`git push origin name-for-your-branch` to save your work online.
+
+7.Open a Pull Request: To create one, go to your fork of the project, click on the Pull Requests tab, and click the big green “New Pull Request” button. After you choose your branch,
```
mallow111, line 43: I would encourage people to link the Issue url in the PR or create one if it is a new issue

mengyaoyang11: that's a good point, thanks Min!

CONTRIBUTING.md
```diff
@@ -1,14 +1,64 @@
...
+It’s a good idea to comment on an Issue that you want to help with and ask some questions before doing any major coding work
+
+2.Fork the repository: Look for the “Fork” button in the top right corner of [salesforce/sloop repo](https://github.com/salesforce/sloop)
+
+3.Clone the fork: Click on the green “clone or download” button,you can choose “clone with HTTPS.” Or, you can use SSH, which is compatible with 2-factor authentication, but just [some more work to set up](https://docs.github.com/en/github/authenticating-to-github/connecting-to-github-with-ssh).
+Go to the folder where you want to save the repository, type `git clone that-https-url-you-copied`
+
+4: Add a remote pointing to the original repository so you can sync changes: `git remote add upstream git@github.com:salesforce/sloop.git`
+
+
```
nurland, line 21: nit: remove extra blank line?

nurland (dismissed):

CONTRIBUTING.md
```diff
@@ -1,14 +1,63 @@
...
+Go to the folder where you want to save the repository, type `git clone that-https-url-you-copied`
+
+4: Add a remote pointing to the original repository so you can sync changes: `git remote add upstream git@github.com:salesforce/sloop.git`
+
+5.Make a git branch: If you type git status you will see the branch name that you’re on, called master. In most projects, master is a special place where the most stable, reviewed, up-to-date code is. So, you’ll need to make your own branch: 
+
+`git branch a-descriptive-name`
+
+Then switch to that branch:
+`git checkout name-for-your-branch`
```
kritikapradhan, line 26: consider changing to "git checkout -b <branch-name>"

mengyaoyang11: That's a good catch, Thanks Kritika! already updated it

CONTRIBUTING.md
```diff
@@ -1,14 +1,63 @@
...
+2.Fork the repository: Look for the “Fork” button in the top right corner of [salesforce/sloop repo](https://github.com/salesforce/sloop)
+
+3.Clone the fork: Click on the green “clone or download” button,you can choose “clone with HTTPS.” Or, you can use SSH, which is compatible with 2-factor authentication, but just [some more work to set up](https://docs.github.com/en/github/authenticating-to-github/connecting-to-github-with-ssh).
+Go to the folder where you want to save the repository, type `git clone that-https-url-you-copied`
+
+4: Add a remote pointing to the original repository so you can sync changes: `git remote add upstream git@github.com:salesforce/sloop.git`
+
+5.Make a git branch: If you type git status you will see the branch name that you’re on, called master. In most projects, master is a special place where the most stable, reviewed, up-to-date code is. So, you’ll need to make your own branch: 
+
+`git branch a-descriptive-name`
```
nurland, line 23: we don't need this command, the command below creates and switches to a new branch

mengyaoyang11: @nurland I think it's better to keep step5 as two steps: create a new branch as one step and switch to the branch as another step. We definitely can combine them into one command as `git checkout -b name-for-your-branch` but I still think it's more descriptive and clear to keep two steps. FYI @kritikapradhan

CONTRIBUTING.md
mengyaoyang11: @kritikapradhan good catch, `git checkout -b` is used to create AND switch to a new branch, since we've already create a new branch, `git checkout branch_name` is more reasonable to just switch to the branch

CONTRIBUTING.md
nurland: I think most people use just `git checkout -b` instead of 2 longer commands where you need to type branch name twice. But I don't feel strong about it.

mengyaoyang11: Sounds reasonable, already updated.

nurland (approved):

mallow111 (approved):

PR #144, opened 2021-02-18
Set priority for input parameters 

kritikapradhan:
Problem - Previous implementation did not have a level of priority for input config paramters. With a config file specified, we would ignore any other arguments passed in cmd line.
Solution-The way I implemented this, is by method of least changes to meet the functionality. Level of priority in order-
1. Command line args
2. Config file
3. Default values
Tried adding test cases to validate implementation for folks, proved to be harder than i thought(would have to run the entire web application, giving it sample command line args and compare logs), feel free to suggest an easier/quicker method to test


mallow111:
can we have a simple description in the PR to talk about what is the issue and what is the proposed solution so people who dont have context can pick it up quickly too.

pkg/sloop/server/internal/config/config.go
```diff
@@ -129,23 +129,29 @@ func registerFlags(fs *flag.FlagSet, config *SloopConfig) {
...
-	if configFilename != "" {
-		newConfig = loadFromFile(configFilename)
-	} else {
-		registerFlags(flag.CommandLine, newConfig)
-		flag.Parse()
+
+	if configFileName != "" {
+		finalConfig = loadFromFile(configFileName,finalConfig)
 	}
+	//register cmd line args
```
mallow111, line 150: what would happen if configFileFlag == "" and configFileOS =="" and configFileName == ""

pkg/sloop/server/internal/config/config.go
```diff
@@ -25,57 +25,57 @@ import (
...
+	ConfigFile string	`json:"config"`
 	// These fields can only come from file because they use complex types
 	LeftBarLinks  []webserver.LinkTemplate         `json:"leftBarLinks"`
 	ResourceLinks []webserver.ResourceLinkTemplate `json:"resourceLinks"`
 	// Normal fields that can come from file or cmd line
-	DisableKubeWatcher       bool          `json:"disableKubeWatch"`
-	KubeWatchResyncInterval  time.Duration `json:"kubeWatchResyncInterval"`
-	WebFilesPath             string        `json:"webfilesPath"`
-	BindAddress              string        `json:"bindAddress"`
+	DisableKubeWatcher       bool          `json:"disable-kube-watch"`
```
duke-harlan, line 34: does this require changes to helm charts (any place configmaps are stored)?

pkg/sloop/server/internal/config/config.go
kritikapradhan: it would use the default values.

mallow111: My concern would be all of the 3 argus are all "", how are we going to handle it, maybe we can have a log to print out these variables, it can help for debugging purposes.

kritikapradhan: see L133, i register default flags. it will never be empty. We are already printing out the values in server.go.

pkg/sloop/server/internal/config/config.go
kritikapradhan: Yes, you would have to update the config json file with these tags(if you are specifying a config file).

pkg/sloop/server/internal/config/config.go
mallow111: i see

PR #145, opened 2021-02-19
Set priority for input parameters

kritikapradhan:
Problem - Previous implementation did not have a level of priority for input config parameters. With a config file specified, we would ignore any other arguments passed in cmd line.
Solution-The way I implemented this, is by method of least changes to meet the functionality. Level of priority in order-
1. Command line args
2. Config file
3. Default values

Tried adding test cases to validate implementation for folks, proved to be harder than i thought(would have to run the entire web application, giving it sample command line args and compare logs), feel free to suggest an easier/quicker method to test


pkg/sloop/server/internal/config/config.go
```diff
@@ -75,77 +75,130 @@ type SloopConfig struct {
...
 // If not there, check the environment variable
 // If we have a config path, load initial values from it
 // Next parse flags again and override any fields from command line
 //
 // We do this to support settings that can come from either cmd line or config file
 func Init() *SloopConfig {
-	newConfig := &SloopConfig{}
+	finalConfig := registerDefaultSloopConfig()
+
+	configFileName:=""
```
nurland, line 183: nit: could you run go fmt, seems like inconsistent spacing here and below

pkg/sloop/server/internal/config/config_test.go
```diff
@@ -16,8 +16,9 @@ func Test_loadFromJSONFile_Success(t *testing.T) {
 	err = json.Unmarshal(configFile, &expectedconfig)
 	assert.Nil(t, err)
 
-	out_config := loadFromFile(configfilename)
-	assert.Equal(t, out_config, &expectedconfig)
+	var outConfig *SloopConfig
+	outConfig = loadFromFile(configfilename,outConfig)
+	assert.Equal(t, outConfig, &expectedconfig)
```
nurland, line 21: can we also add some unit tests that ensure overrides are taking precedence and default values are not changed if they are not present in config/flag.

pkg/sloop/server/internal/config/config.go
```diff
@@ -75,77 +75,130 @@ type SloopConfig struct {
...
 	}
-	if configFilename != "" {
-		newConfig = loadFromFile(configFilename)
-	} else {
-		registerFlags(flag.CommandLine, newConfig)
-		flag.Parse()
+	if configFileName != "" {
+		finalConfig = loadFromFile(configFileName,finalConfig)
 	}
+	finalConfig=registerFlags(flag.CommandLine,finalConfig)
```
nurland, line 197: I don't think we need to return config from registerFlags.

nurland (changes requested):
Left some comments, but overall looks good.

pkg/sloop/server/internal/config/config.go
garakly: +1

pkg/sloop/server/internal/config/config.go
```diff
@@ -75,77 +75,130 @@ type SloopConfig struct {
...
 // Next parse flags again and override any fields from command line
 //
 // We do this to support settings that can come from either cmd line or config file
 func Init() *SloopConfig {
-	newConfig := &SloopConfig{}
+	finalConfig := registerDefaultSloopConfig()
+
+	configFileName:=""
+	configFileFlag := preParseConfigFlag()
+	configFileOS := os.Getenv(sloopConfigEnvVar)
```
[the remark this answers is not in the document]
kritikapradhan: it's for the logging purposes, wanted to log if we are using the env var or the flag passed. It adds clarification and answers which config file we prioritize, in case both are set.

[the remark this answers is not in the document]
nurland: with the snippet above we can also do the same logging, can't we?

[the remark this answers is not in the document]
kritikapradhan: Reusing variables adds an element of risk, making the code harder to follow and more error prone.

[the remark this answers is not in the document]
nurland: IMO, right now the code is less readable than snippet above, harder to follow. This adds complexity and can also be error prone.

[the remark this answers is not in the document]
kritikapradhan: I don't agree it's less readable currently. but we can get more opinions, see what others think

pkg/sloop/server/internal/config/config_test.go
```diff
@@ -16,8 +19,9 @@ func Test_loadFromJSONFile_Success(t *testing.T) {
 	err = json.Unmarshal(configFile, &expectedconfig)
 	assert.Nil(t, err)
 
-	out_config := loadFromFile(configfilename)
-	assert.Equal(t, out_config, &expectedconfig)
+	var outConfig *SloopConfig
+	outConfig = loadFromFile(configfilename, outConfig)
+	assert.Equal(t, outConfig, &expectedconfig)
```
garakly, line 24: assert package expects first the `expected` value to be passed, and then the `actual` value.
Please fix what is being passed as expected, and what is passed as actual.

Documentation: 
> Equal func(t TestingT, expected interface{}, actual interface{}, msgAndArgs ...interface{}) bool

pkg/sloop/server/internal/config/config_test.go
```diff
@@ -27,16 +31,28 @@ func Test_loadFromYAMLFile_Success(t *testing.T) {
 	err = yaml.Unmarshal(configFile, &expectedconfig)
 	assert.Nil(t, err)
 
-	out_config := loadFromFile(configfilename)
-	assert.Equal(t, out_config, &expectedconfig)
+	var outConfig *SloopConfig
+	outConfig = loadFromFile(configfilename, outConfig)
+	assert.Equal(t, outConfig, &expectedconfig)
```
garakly, line 36: same as above comment regarding `assert` package.

pkg/sloop/server/internal/testFiles/testconfig.yaml
```diff
@@ -1,43 +1,14 @@
 ---
 ConfigFile: ''
 leftBarLinks: 
-resourceLinks:
```
garakly, line 4: Why are we removing these values from the test configs?

pkg/sloop/server/internal/testFiles/testconfig.json
```diff
@@ -2,21 +2,6 @@
 	"ConfigFile": "",
 	"leftBarLinks": null,
 	"resourceLinks": null,
-	"disableKubeWatch": true,
```
garakly, line 5: Why are we removing these values from the test configs?

garakly (changes requested):

pkg/sloop/server/internal/config/config.go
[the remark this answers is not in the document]
garakly: Ok, so two things:
1) Yes, you can definitely add the logs to the snippet I proposed, no worries there. Here's what that can look like:

``` bash
  # Set config file name
	configFileName :=  preParseConfigFlag()
	if configFileName == ""  {
		configFileName := os.Getenv(sloopConfigEnvVar)
                glog.Infof("Set via environmental variables: %s", configFileName)
	} else {
               glog.Infof("Set via flag: %s", configFileName)
        }

# Load config file
        if configFileName != "" {
                 finalConfig = loadFromFile(configFileName, finalConfig)
        }
```

2) I am with nurland on this - I **do not want** to know about the different ways of setting the configFileName - all that we need, is a way to determine how to set it. Once set - we use it. That makes the proposed code snippet easier to read. 

With multiple variables, now I need to know **which** ones are set, **why** they are set, and **what** I need to do with them - individually. The proposed snippet says - set the config file name either first from the flags, or second, from the env vars, then load it.

pkg/sloop/server/internal/config/config_test.go
garakly: +1

pkg/sloop/server/internal/config/config.go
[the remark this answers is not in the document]
kritikapradhan: I don't understand why both of u are against defining variables? I don't think it is harming either the functionality or the readability. The implementation currently is basically "lets get all the information on the table and then decide which to use" rather than "let me get some information and see if that can be used, if not let me try something else". 
The only actual difference I see between current and proposed implementation, is **two** variables.

pkg/sloop/server/internal/config/config_test.go
kritikapradhan: any pointers on what way would be acceptable to u?

pkg/sloop/server/internal/testFiles/testconfig.yaml
kritikapradhan: to check if it works the way we want with an incomplete config file input.

garakly: gotcha, thanks for the info 👍

pkg/sloop/server/internal/config/config.go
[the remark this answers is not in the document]
nurland: I think my main point is that _for me_ the current code was much harder to follow than snippet above. Seems like it was hard to follow for @garakly as well. Having said that I don't feel strongly about, not a PR blocker for me.

pkg/sloop/server/internal/config/config_test.go
nurland: any way that covers the new behavior (flags>config>default value).

pkg/sloop/server/internal/config/config.go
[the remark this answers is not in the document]
kritikapradhan: makes sense! let me try to make it more readable.

nurland (dismissed):
Left minor comment, otherwise LGTM

nurland (approved):

garakly (approved):

PR #146, opened 2021-03-02
Add User Input End Time as a Parameter

mengyaoyang11:
Problem:
Custom Time Search enables user to select an end time, right now we have end time defaulted to `now`. Need to add a configurable end time parameter and pass it to the existing query.

Solution:
Add a new parameter `userInputEndTime`, compared to the output of function `getEndOfTime`, the earlier one will be the actual end time.

### DIFF
--- a/pkg/sloop/queries/timerange.go
+++ b/pkg/sloop/queries/timerange.go
@@ -43,9 +43,7 @@ func computeTimeRange(params url.Values, tables typed.Tables, maxLookBack time.D
 // Compare result of function `getEndOfTime` and user defined end time,
 // return the earlier one
 func computeActualEndTime(userInputEndTime string, endOfTime time.Time) time.Time {
-	var userDefinedEndOfTime time.Time
-	var err error
-	userDefinedEndOfTime, err = parseUnixTimeString(userInputEndTime)
+	userDefinedEndOfTime, err := parseUnixTimeString(userInputEndTime)
 	if err != nil {
 		glog.Errorf("Invalid EndTimeParam input: %v.  err: %v", userInputEndTime, err)
 		return endOfTime
--- a/pkg/sloop/queries/timerange.go
+++ b/pkg/sloop/queries/timerange.go
@@ -36,7 +36,24 @@ const minLookback = 1 * time.Minute
 // TODO: If wall clock is in the middle of the newest partition min-max time we can use it
 func computeTimeRange(params url.Values, tables typed.Tables, maxLookBack time.Duration) (time.Time, time.Time, error) {
 	endOfTime := getEndOfTime(tables)
-	return computeTimeRangeInternal(params, endOfTime, maxLookBack)
+	actualEndTime := computeActualEndTime(params.Get(EndTimeParam), endOfTime)
+	return computeTimeRangeInternal(params, actualEndTime, maxLookBack)
+}
+
+// Compare result of function `getEndOfTime` and user defined end time,
+// return the earlier one
+func computeActualEndTime(userInputEndTime string, endOfTime time.Time) time.Time {
+	var userDefinedEndOfTime time.Time
+	var err error
+	userDefinedEndOfTime, err = parseUnixTimeString(userInputEndTime)
+	if err != nil {
+		glog.Errorf("Invalid EndTimeParam input: %v.  err: %v", userInputEndTime, err)
+		return endOfTime
+	}
+	if endOfTime.After(userDefinedEndOfTime) {
+		return userDefinedEndOfTime
+	}
+	return endOfTime
 }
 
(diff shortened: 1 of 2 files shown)


pkg/sloop/queries/timerange.go
```diff
@@ -36,7 +36,24 @@ const minLookback = 1 * time.Minute
...
-	return computeTimeRangeInternal(params, endOfTime, maxLookBack)
+	actualEndTime := computeActualEndTime(params.Get(EndTimeParam), endOfTime)
+	return computeTimeRangeInternal(params, actualEndTime, maxLookBack)
+}
+
+// Compare result of function `getEndOfTime` and user defined end time,
+// return the earlier one
+func computeActualEndTime(userInputEndTime string, endOfTime time.Time) time.Time {
+	var userDefinedEndOfTime time.Time
+	var err error
```
kritikapradhan, line 47: consider removing both declarations and change L48 to userdefinedendoftime,err **:=** parseUnixtimestring().

pkg/sloop/queries/timerange_test.go
```diff
@@ -73,6 +73,25 @@ func Test_timeRangeTable(t *testing.T) {
...
+	userInputEndTime = fmt.Sprintf("%v", someQueryEndTs.UTC().Unix())
+	// endOfTIme is before userInput
+	result = computeActualEndTime(userInputEndTime, endOfTime)
+	assert.Equal(t, result, endOfTime)
+
+	// endOfTIme is after userInput
+	endOfTime = someQueryEndTs.Add(time.Hour*2)
+	result = computeActualEndTime(userInputEndTime, endOfTime)
+	assert.Equal(t, result, someQueryEndTs)
+
```
kritikapradhan, line 89: can we have a test for if they're equal ?

pkg/sloop/queries/timerange.go
mengyaoyang11: Thanks for the suggestion and knowledge sharing @kritikapradhan Updated as suggested

pkg/sloop/queries/timerange_test.go
mengyaoyang11: @kritikapradhan That's a good point, added a test case for equal value

pkg/sloop/queries/timerange.go
```diff
@@ -36,7 +36,22 @@ const minLookback = 1 * time.Minute
 // TODO: If wall clock is in the middle of the newest partition min-max time we can use it
 func computeTimeRange(params url.Values, tables typed.Tables, maxLookBack time.Duration) (time.Time, time.Time, error) {
 	endOfTime := getEndOfTime(tables)
-	return computeTimeRangeInternal(params, endOfTime, maxLookBack)
+	actualEndTime := computeActualEndTime(params.Get(EndTimeParam), endOfTime)
```
nurland, line 39: `computeTimeRangeInternal` already takes into account `params.Get(EndTimeParam)`, not sure what we are trying to do here.

nurland (changes requested):

pkg/sloop/queries/timerange.go
mengyaoyang11: @nurland endOfTime is `now`, params.Get(EndTimeParam) is user defined end time, we want to compare which one is earlier, and take the earlier one as the actual end time. Please see the computeActualEndTime function below, which explains how to determine the actualEndTime

nurland: @mengyaoyang11 endOfTime is the latest available partition in badgerDb.
If you look at he following lines, we already account for user defined end time and we only override it when computedEnd (based on user input) is beyond endOfTime(latest partition) 
https://github.com/salesforce/sloop/pull/146/files#diff-3f5eeba9668570e88b8263b0d459cc5836ed590fde2bacfcd85a25b30ec3973eR88
https://github.com/salesforce/sloop/pull/146/files#diff-3f5eeba9668570e88b8263b0d459cc5836ed590fde2bacfcd85a25b30ec3973eR95

nurland: That been said - we only do that when we pass startTime and endTime, when we pass LookbackInterval we set computedEnd to the endOfTime. I suggest we make change in https://github.com/salesforce/sloop/pull/146/files#diff-3f5eeba9668570e88b8263b0d459cc5836ed590fde2bacfcd85a25b30ec3973eR81 and keep endOfTime intact

sana-jawad: yes agreed this would be much cleaner. So in summary it would be:
1. Keep endOftime intact
2. Make change in computeTimeRangeInternal function where the endTime is set to endOfTime

mengyaoyang11: @nurland The links in you review directs me to the diff files of my PR, instead of the code lines you referred to, could you please repost it so I known which lines of code you are talking about? Thanks!

PR #147, opened 2021-03-04
Add User Input End Time

mengyaoyang11:
Problem:
Custom Time Search new UI has end time set either as default now or as customer input. 

Solution:
Update computeTimeRangeInternal function to make sure no error return when endTimeVal and lookBackVal are given

Note:
1) keep startTimeVal as a parameter for now even though its not exposed to both current/new UI, will totally remove startTimeVal once we are sure its not used anywhere in new feature.
2) Keep the input validation for now even though the new feature ensures lookBackVal will not be empty.

### DIFF
--- a/pkg/sloop/queries/timerange.go
+++ b/pkg/sloop/queries/timerange.go
@@ -53,8 +53,8 @@ func computeTimeRangeInternal(params url.Values, endOfTime time.Time, maxLookBac
 		return time.Time{}, time.Time{}, fmt.Errorf("Time range must be set with either [%v] or both of [%v,%v] but all 3 were empty", LookbackParam, StartTimeParam, EndTimeParam)
 	}
 	if lookBackVal != "" {
-		if startTimeVal != "" || endTimeVal != "" {
-			return time.Time{}, time.Time{}, fmt.Errorf("When [%v] is set, you can not set either of [%v,%v].  Got (%v,%v,%v) respectively", LookbackParam, StartTimeParam, EndTimeParam, lookBackVal, startTimeVal, endTimeVal)
+		if startTimeVal != "" {
+			return time.Time{}, time.Time{}, fmt.Errorf("When [%v] is set, you can not set both of [%v,%v] or set only [%v].  Got (%v,%v,%v) respectively", LookbackParam, StartTimeParam, EndTimeParam, StartTimeParam, lookBackVal, startTimeVal, endTimeVal)
 		}
 	} else {
 		if (startTimeVal == "") || (endTimeVal == "") {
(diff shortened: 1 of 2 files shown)


pkg/sloop/queries/timerange.go
```diff
@@ -62,18 +62,18 @@ func computeTimeRangeInternal(params url.Values, endOfTime time.Time, maxLookBac
 		}
 	}
 
+	// always use customer input as computedEnd
+	computedStart, computedEnd, err = getTimeRangeFromStartEnd(startTimeVal, endTimeVal)
```
nurland, line 66: @mengyaoyang11 we are breaking the existing behavior - if lookBack is passed then startTime and endTime will be empty and getTimeRangeFromStartEnd will return error.

nurland (changes requested):

pkg/sloop/queries/timerange_test.go
```diff
@@ -55,7 +59,8 @@ func Test_timeRangeTable(t *testing.T) {
 		{"", "abc", "123", true, time.Time{}, time.Time{}},
 		{"", "123", "abc", true, time.Time{}, time.Time{}},
 	}
-	for _, thisRange := range rangeTests {
+	for i, thisRange := range rangeTests {
+		fmt.Print(i)
```
nurland, line 63: why do we need this? can we remove it?

mengyaoyang11: It helps especially when test case failed. With the number printed out you can easily target which test is failed. Of course we can remove it does not make sense.

pkg/sloop/queries/timerange.go
mengyaoyang11: updated

pkg/sloop/queries/timerange_test.go
nurland: Can we just use t.Run arg for that? E.g. we can update the line below to be something like this:

t.Run(fmt.Sprintf("Test id: %v, range: %+v", i, thisRange)

mengyaoyang11: updated

pkg/sloop/queries/timerange_test.go
```diff
@@ -55,8 +59,8 @@ func Test_timeRangeTable(t *testing.T) {
 		{"", "abc", "123", true, time.Time{}, time.Time{}},
 		{"", "123", "abc", true, time.Time{}, time.Time{}},
 	}
-	for _, thisRange := range rangeTests {
-		t.Run(fmt.Sprintf("%+v", thisRange), func(t *testing.T) {
+	for i, thisRange := range rangeTests {
+		t.Run(fmt.Sprintf("Test id: %v, thisRange: %v", i, thisRange), func(t *testing.T){
```
nurland, line 63: nit: just curious why did we change `%+v` to `%v`?

mengyaoyang11: you are correct, change back to `%+v` as plus flag adds field names

nurland (approved):

PR #148, opened 2021-04-01
Add custom time window which enables users to select the end time

mengyaoyang11:
Problem:
Custom Time Search new UI should have a new window which enables user to select end time

Solution:
Add custom time window for end time. 
The ui will render such behavior:
1) The default of the end time is now, after the page refreshes or a submit is clicked(query is sent to backend), the date on ui will be 'now', which is expected behavior.
2) The max allowed user selected time is set to now to prevent users enter invalid date.

UI screenshot
Screen Shot 2021-04-01 at 9 14 16 AM


mengyaoyang11:
Close it for now, need to wait till it functionally work

PR #150, opened 2021-04-27
Add custom time search window

mengyaoyang11:
Problem: Currently Sloop only lets you pick the duration of the timeline ending roughly "now". Would like the ability to change the start time and end time. This is useful when investigating something that happened in the past but over a short time duration.

Solution: 
1. Use a native datetime-local element which enables user to select a time range they preferred. 
2. Use sessionStorage to store user selected end time
3. Get user selected end time from session storage and pass it to backend handler
4. User selected end time will stay on ui after click submit button or refresh the page
5.  If current window is shut down and a new window is open, the default time on ui is 'now'

One thing needs to improve: The native datetime-local element didn't support display both user selected time and 'now' on ui: tried many ways but default now is always overwritten by user selected time. So create another follow-up WI to explore possibility to set default 'now' either with this element or using new element


pkg/sloop/webserver/webfiles/sloop_ui.js
```diff
@@ -97,6 +97,7 @@ function render(result) {
 
 
     if (!data) {
+        console.log("!!!!!!!!!! inside no data !!!!!")
```
nurland, line 100: remove?

pkg/sloop/queries/timerange.go
```diff
@@ -167,6 +165,16 @@ func parseUnixTimeString(unixStr string) (time.Time, error) {
 	return time.Unix(unixNum, 0).UTC(), nil
 }
 
+func parseTimestampString(str string) (time.Time, error) {
```
sana-jawad, line 168: please add unit tests for this function

pkg/sloop/queries/timerange.go
```diff
@@ -167,6 +165,16 @@ func parseUnixTimeString(unixStr string) (time.Time, error) {
...
+	// Golang constant value, do NOT change
+	layout := "2006-01-02T15:04:05"
```
sana-jawad, line 170: use const keyword to declare constant

pkg/sloop/webserver/webfiles/filter.js
```diff
@@ -75,6 +75,16 @@ function populateDropdownFromQuery(param, elementId, defaultValue, queryUrl) {
     return value
 }
 
+var selectedEndTimeValue;
+$(document).ready(function(e) {
+    $('#selectedEndTime').on('blur', function(evt) {
+        var dateTimeValue = document.getElementById('selectedEndTime').value;
+        selectedEndTimeValue = dateTimeValue.toString()
+        sessionStorage.setItem('selectedEndTime', selectedEndTimeValue);
+        sessionStorage.setItem('setSelectedEndTime', new Date().toISOString());
```
sana-jawad, line 84: Whats the purpose of this variable?

pkg/sloop/webserver/webfiles/sloop_ui.js
sana-jawad, line 100: please remove

pkg/sloop/webserver/webfiles/filter.js
mengyaoyang11: this variable is basically used to keep track of the time when user pick an end time, it's used here: https://github.com/salesforce/sloop/pull/150/files#diff-0d1cf3a5e1a60113eebad58ab49d836c329c820390474fcedc837693619757c9R533,having this variable, we make sure user has a selection and we then keep users selection and render in ui

nurland: @mengyaoyang11 why do we need `selectedEndTimeValue` var declared at the global scope, I don't see it is used anywhere except line 83, which it seems can be replaced with dateTimeValue.toString(), i.e.:
`sessionStorage.setItem('selectedEndTime', dateTimeValue.toString());`
if it is for readability/debugging - can we reduce the scope (same as `var dateTimeValue`)?

pkg/sloop/queries/timerange.go
mengyaoyang11: Thanks, tests added

pkg/sloop/webserver/webfiles/filter.js
mengyaoyang11: @nurland thanks, good catch, moved selectedEndTimeValue inside the function

pkg/sloop/webserver/webfiles/sloop_ui.js
mengyaoyang11: removed

pkg/sloop/queries/timerange.go
mengyaoyang11: updated

pkg/sloop/webserver/webfiles/index.html
```diff
@@ -16,6 +16,9 @@
     <title>Sloop{{if (ne .CurrentContext "")}} - {{.CurrentContext}}{{end}}</title>
     <link rel="stylesheet" type="text/css" href="webfiles/sloop.css">
     <link rel='shortcut icon' type='image/x-icon' href='webfiles/favicon.ico'/>
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap@4.5.3/dist/css/bootstrap.min.css"
+          integrity="sha384-TX8t27EcRE3e/ihU7zmQxVncDAy5uIKz4rEkgIXeMed4M0jlfIDPvg6uqKI2xXr2"
```
sana-jawad, line 20: As there are UI changes you would need to update pkg/sloop/webserver/bindata.go  to update binary data. Please refer to this PR https://github.com/salesforce/sloop/pull/138/files and ReadMe https://github.com/salesforce/sloop#updating-webfiles-folder

pkg/sloop/webserver/webfiles/sloop_ui.js
```diff
@@ -514,4 +514,26 @@ function showDetailedTooltip(d, event, parent) {
...
+    const now = new Date();
+    var iso = now.toISOString();
+    var maxDate = iso.substring(0,iso.length-5);
+    elem = document.getElementById("selectedEndTime");
+    elem.value = maxDate;
+    elem.max = maxDate;
+
+    //Display user selected end time on ui after click submit button or refresh the page
+    var userDate = new Date();
+    // check if selected end time happened within 3 seconds
```
sana-jawad, line 530: Why do we value 3 sec here?

mengyaoyang11: 3 sec is a random reasonable number, any short period is okay

pkg/sloop/webserver/webfiles/index.html
mengyaoyang11: @sana-jawad Thanks for the info, I noticed in readme stating run 'go-bindata -o bindata.go webfiles/' before submit a PR, I am not quite sure if it matters if I run it now(PR already submitted), and all I need to do is to include the updated bindata.go file in the PR, right?

sana-jawad: You can include the updated binddata.go in this PR

mengyaoyang11: Thanks @sana-jawad. Updated: include bindata.go in the PR

pkg/sloop/webserver/bindata.go
```diff
@@ -15,7 +15,7 @@
 // webfiles/sloop_ui.js
 // DO NOT EDIT!
 
-package webserver
```
sana-jawad, line 18: Why do we have the package name change?

sana-jawad (commented):
Just added one more comment. Overall it looks good to me

pkg/sloop/webserver/bindata.go
mengyaoyang11: I have no idea why its changed, all I did is to run 'go-bindata -o bindata.go webfiles/' within the webserver directory. Can I manually change/update the package name in bindata.go file? Or I should run the command again?

mengyaoyang11: @sana-jawad Manually change package name back to webserver

sana-jawad (approved):

PR #152, opened 2021-05-06
enable user reset date/time to now

mengyaoyang11:
Problem: Need to provide customer with ability to select now as a reset value
Solution: Create a button and each click on the button will reset the time to now

Screen Shot 2021-05-06 at 11 33 28 AM


pkg/sloop/webserver/webfiles/sloop_ui.js
```diff
@@ -536,4 +536,13 @@ $(document).ready(function() {
     userDate.setMilliseconds(null);
     document.getElementById('selectedEndTime').value = userDate.toISOString().slice(0, -1);
 
+    $('#now').click(function(){
+        const resetNow = new Date();
+        resetNow.setMinutes(resetNow.getMinutes() - resetNow.getTimezoneOffset());
+        resetNow.setMilliseconds(null);
+        document.getElementById('selectedEndTime').value = resetNow.toISOString().slice(0, -1);
+        sessionStorage.clear();
```
duke-harlan, line 544: does `sessionStorage` contain any values other then `selectedEndTime`? If so, `clear()` may be destructive

pkg/sloop/webserver/webfiles/index.html
```diff
@@ -36,9 +36,10 @@
             <select name="query" id="filterquery" hidden="true"> </select>
             <!-- <br><br> -->
 
-            <label for="selectedEndTime">End time</label><br/>
+            <label for="selectedEndTime">End Time</label><br/>
             <div>
                 <input type="datetime-local" step="1" id="selectedEndTime" name="selectedEndTime" />
+                <input type="button" value="now" name="now" id="now">
```
sana-jawad, line 42: nit: Can you please change label to "Now"

duke-harlan (commented):
other then the scope of `clear()` things look ok

pkg/sloop/webserver/webfiles/sloop_ui.js
mengyaoyang11: Yeah, We have another val called `setSelectedEndTime`, which is used to keep track when user select a time on date time picker. setSelectedEnd Time is set [here](setSelectedEndTime) and used [here](setSelectedEndTime). If clear() is a concern, it can be fixed by using remove() instead

pkg/sloop/webserver/webfiles/index.html
mengyaoyang11: done

pkg/sloop/webserver/webfiles/sloop_ui.js
mengyaoyang11: updated by using `removeItem()` other than `clear()`

duke-harlan (approved):
lgtm - I'd prefer if someone with more UI experience also sign off

sana-jawad (dismissed):

sana-jawad:
The changes look good to me. But it looks like the build is failing due to a unit test failure since your last PR. If you want to send a separate PR to fix it, we can merge this one @mengyaoyang11

mengyaoyang11:
@sana-jawad Updated test file to fix the build failure. Please review, thanks
Screen Shot 2021-05-10 at 9 03 17 PM

PR #156, opened 2021-05-20
Fix end time

tyrken:
With current master the "Details" view on selecting a resource was broken, with errors in the server log about missing "endtime" param ... as the golang code is using `end_time` in https://github.com/salesforce/sloop/blob/14ae23c068c64c25ccaa586d4e6bfff728bda846/pkg/sloop/webserver/resourcehandler.go#L86-L90

Hope you don't mind @mengyaoyang11 @sana-jawad but I'm reverting #151 to keep the "xxx_time" variables all in the same style with each other, and updated what I hope is one bit of JS code that uses it.

Also updated the Readme to include a better form of go-bindata command which is needed to keep the package name `webserver` without manual fixup, at least with the v3.1.2 I have...


mengyaoyang11:
Hi @tyrken Thanks for the insight and contribution, I am wondering if we can update the param end_time to endtime [here](https://github.com/salesforce/sloop/blob/14ae23c068c64c25ccaa586d4e6bfff728bda846/pkg/sloop/webserver/resourcehandler.go#L86-L9)  instead of reverting the changes? The reason why I suggest that is because reverting the changes won't return correct results for custom time search function. You can try on your side by reverting the changes and most likely interacting with custom time search window gives you incorrect page. When I say Custom time search function I mean the date time picker in the following screenshot. It does not return the data for the time frame you pick if reverting the changes. 
117369929-7bba9280-ae7a-11eb-9903-b68af9120517


By the way, I appreciate your update on go-bindata command, totally agree on that change.

tyrken:
Thanks for the quick response @mengyaoyang11 , but the date picker (shown when clicking on the small calendar icon next to the "Now" button) work for me with this PR ... as I updated line 119 to create "end_time" also & regenerated the go-bindata.

Is there some other code somewhere else which is forcing the variable to be named differently & can't be changed?  Or are you talking about some different UI component I haven't discovered yet?

mengyaoyang11:
@tyrken oh that's surprised to know, just want to confirm: with your change when you select a specific end time, and pick a look back range, the query returns the exact data for the time range you pick? If that is the case, then I will say let's go ahead with your changes. But before approve your changes, it would be great if you can have a screenshot to show it returns the correct query. I hope it should not be that complex

tyrken:
I double checked and I think it is indeed working as hoped - you can use the picker to set a date/time easily (first screenshot): 
Date-picker-setting-end-time

... and the resulting view (after hitting Submit) shows according to requested end & look-back-duration (second screenshot, look-back was 1 hour):
Data-returned-to-times

mengyaoyang11 (approved):
LGTM

sana-jawad (approved):

PR #160, opened 2021-06-10
Removing end time temporarily

akrag:


pkg/sloop/webserver/webfiles/index.html
```diff
@@ -36,11 +36,13 @@
             <select name="query" id="filterquery" hidden="true"> </select>
             <!-- <br><br> -->
 
+            <!-- Temporarily disable Endtime (Workaround for W-9423151)
```
sana-jawad, line 39: As this is UI change you would need to update binddata.go file too

akrag: Updated webfiles folder

sana-jawad (approved):

PR #161, opened 2021-06-14
Present User Selected Time and Now in UTC format

mengyaoyang11:
Problem statement:
Currently clicking the "Now" button sends the local time now to the query, not the UTC time, since there is a time difference between local time(either in the US or in India) and UTC, if the sloop instance is not running during that time period, we will get an empty screen after clicking the 'Now' button. This is expected behavior, customer who keeps running sloop for a long time will not have this issue.

Solution:
Display both customer selected time and now in UTC format


pkg/sloop/webserver/webfiles/sloop_ui.js
```diff
@@ -532,13 +533,13 @@ $(document).ready(function() {
         && ! Date.parse(sessionStorage.getItem('setSelectedEndTime')) < new Date(userDate.getTime() - 3000)){
         userDate = new Date(sessionStorage.getItem('selectedEndTime'));
     }
-    userDate.setMinutes(now.getMinutes() - now.getTimezoneOffset());
+    userDate.setMinutes(userDate.getMinutes() - userDate.getTimezoneOffset());
```
duke-harlan, line 536: does `.getUTCMinutes()` (as seen line 542) work here?

duke-harlan (approved):
it looks like we're using 2 techniques to accomplish the same change - if they produce the same behavior, lets stick with just one

pkg/sloop/webserver/webfiles/sloop_ui.js
mengyaoyang11: Tried `.getUTCMinutes()` here as well but not producing the same behavior. Will talk to Sana to see if the change is needed or not

mengyaoyang11: @duke-harlan Updated the pr to address the comment: use one technique to present both user selected time and now to local time. Please review again

duke-harlan (dismissed):

sana-jawad (changes requested):
Please make sure the same time zone is used across.

sana-jawad (approved):

PR #171, opened 2021-09-01
Unrelated Pod Information

kritikapradhan:
Problem-

A customer has a multiple deployments with names for eg. deployment1, deployment11, deployment12
Sloop would mismatch information and display information for deployment11 in deployment1

Solution -
The prefix string used to match the payloads was missing a forward slash. Without this, it would add the others because the prefix would match.

### DIFF
--- a/pkg/sloop/queries/respayloadquery_test.go
+++ b/pkg/sloop/queries/respayloadquery_test.go
@@ -182,3 +182,33 @@ func Test_removeDupePayloads_twoTheSame_returnsFirstOne(t *testing.T) {
 	ret := removeDupePayloads(input)
 	assert.Equal(t, expected, ret)
 }
+
+func Test_GetResPayload_True_HasSamePrefix(t *testing.T) {
+	untyped.TestHookSetPartitionDuration(time.Hour)
+	partitionId := untyped.GetPartitionId(someTs)
+	values := helper_get_params()
+	expectedKind := "someKind"
+	expectedNS := "someNamespace"
+	expectedName := "someName-11"
+	values[KindParam] = []string{expectedKind}
+	values[NamespaceParam] = []string{expectedNS}
+	values[NameParam] = []string{expectedName}
+
+	var keys []string
+	keys = append(keys, typed.NewWatchTableKey(partitionId, expectedKind, expectedNS, expectedName, someTs).String())
+	keys = append(keys, typed.NewWatchTableKey(partitionId, "someKind", "someNamespace", "someName-1", someTs).String())
+	keys = append(keys, typed.NewWatchTableKey(partitionId, "someKind", "someNamespace", "someName-16", someTs).String())
+	keys = append(keys, typed.NewWatchTableKey(partitionId, "someKind", "someNamespace", "someName-15", someTs).String())
+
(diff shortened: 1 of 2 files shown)


duke-harlan:
please add some unit tests that expose the bad behavior and show it no longer works that way

pkg/sloop/queries/respayloadquery_test.go
```diff
@@ -182,3 +182,33 @@ func Test_removeDupePayloads_twoTheSame_returnsFirstOne(t *testing.T) {
...
+	keys = append(keys, typed.NewWatchTableKey(partitionId, "someKind", "someNamespace", "someName-16", someTs).String())
+	keys = append(keys, typed.NewWatchTableKey(partitionId, "someKind", "someNamespace", "someName-15", someTs).String())
+
+	tables := helper_get_resPayload(keys, t, somePTime)
+	res, err := GetResPayload(values, tables, someTs.Add(-1*time.Hour), someTs.Add(6*time.Hour), someRequestId)
+	assert.Nil(t, err)
+	expectedRes := `[
+ {
+  "payloadTime": 1546398245000000006,
+  "payload": "{\n  \"metadata\": {\n    \"name\": \"someName\",\n    \"namespace\": \"someNamespace\",\n    \"uid\": \"6c2a9795-a282-11e9-ba2f-14187761de09\",\n    \"creationTimestamp\": \"2019-07-09T19:47:45Z\"\n  }\n}",
```
sana-jawad, line 209: nit: shouldn't the name here be someName-1?

sana-jawad (approved):

pkg/sloop/queries/respayloadquery_test.go
```diff
@@ -182,3 +182,33 @@ func Test_removeDupePayloads_twoTheSame_returnsFirstOne(t *testing.T) {
...
+	expectedKind := "someKind"
+	expectedNS := "someNamespace"
+	expectedName := "someName-11"
+	values[KindParam] = []string{expectedKind}
+	values[NamespaceParam] = []string{expectedNS}
+	values[NameParam] = []string{expectedName}
+
+	var keys []string
+	keys = append(keys, typed.NewWatchTableKey(partitionId, expectedKind, expectedNS, expectedName, someTs).String())
+	keys = append(keys, typed.NewWatchTableKey(partitionId, "someKind", "someNamespace", "someName-1", someTs).String())
```
duke-harlan, line 199: please use named constant instead of literals: "someKind", "someNamespace". Just use the literals for the new names

pkg/sloop/queries/respayloadquery_test.go
duke-harlan: not a nit - it should match the NameParam inserted in the `values` at line 195

kritikapradhan:
ouch, my internet loaded ur comments after i merged

pkg/sloop/queries/respayloadquery_test.go
kritikapradhan: So that is because the sample payload used for all test cases is defined [here](https://github.com/salesforce/sloop/blob/master/pkg/sloop/queries/respayloadquery_test.go#L22) . The payload key is what needs to be correct in this case

PR #172, opened 2021-09-02
Fix Details dialog with no payload entries

duke-harlan:
Problem: sometimes clicking on an object's history bar brings up a Detail dialog with an empty set of Payloads instead of at least one

The root of the problem is that the method `getSeekKey()` would return a pointer to the source key instead of a duplicate and would mutate the source key.

In GetResPayload(), after the RangeRead, the implementation goes looking for the closest, older key and adds the corresponding payload to the result - this was to ensure the collection always has at least one entry.

To find the closest, older payload, a seekKey is generated from the keyComparator and both are passed in to GetPreviousKey(). The seekKey is mutated with the partition Id while walking backward in time. Each partition is searched by getLastMatchingKeyInPartition() using the seekKey and comparing the found keys against the keyComparator as a prefix.
The keyComparator is constructed without a timestamp. When converted to a string the timestamp is left off. Using this for a prefix match against the database keys.

When seekKey & keyComparator are referring to the same object, the time stamp in seekKey is also part of keyComparator. Then the string form of keyComparator contains a timestamp and never matches against any key coming out of the database.

I also added some verbose logs to expose what is happening for each call to `GetResPayload`
Further testing has not shown an empty payload list (1 corner case - see below)

corner case - early in the Sloop process life when the history of the data is short, if the UI is showing many hours (I was using 6) an existing object is presented as a bar filling left to right of the entire field. The implies that Sloop has data for the entire presented time range. Clicking on an older timestamp for the object may still see 0 payloads. The GetPreviousKey() logic explained above has no previous partitions to compare against.

### DIFF
--- a/pkg/sloop/queries/eventquery_test.go
+++ b/pkg/sloop/queries/eventquery_test.go
@@ -91,7 +91,7 @@ func Test_GetEventData_True(t *testing.T) {
 	values := helper_get_params()
 	values[KindParam] = []string{"someKinda"}
 	values[NamespaceParam] = []string{"someNamespace"}
-	values[NameParam] = []string{"someName"}
+	values[NameParam] = []string{"someName.xx"}
 	var keys []string
 	keys = append(keys, typed.NewWatchTableKey(partitionId, "Event", "someNamespace", "someName.xx", someTs).String())
 	keys = append(keys, typed.NewWatchTableKey(partitionId, "Event", "someNamespaceb", "someName.xx", someTs).String())
--- a/pkg/sloop/queries/respayloadquery.go
+++ b/pkg/sloop/queries/respayloadquery.go
@@ -21,6 +21,10 @@ import (
 	"time"
 )
 
+const (
+	glogVerbose = 10
+)
+
 type ResPayLoadData struct {
 	PayloadList []PayloadOuput `json:"payloadList"`
 }
@@ -32,6 +36,8 @@ type PayloadOuput struct {
 }
 
 func GetResPayload(params url.Values, t typed.Tables, startTime time.Time, endTime time.Time, requestId string) ([]byte, error) {
+
+	glog.V(glogVerbose).Infof("GetResPayload: startTime: %v, endTime: %v", startTime.Unix(), endTime.Unix())
 	var watchRes map[typed.WatchTableKey]*typed.KubeWatchResult
 	var previousKey *typed.WatchTableKey
 	var previousVal *typed.KubeWatchResult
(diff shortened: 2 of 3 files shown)


pkg/sloop/queries/eventquery_test.go
```diff
@@ -91,7 +91,7 @@ func Test_GetEventData_True(t *testing.T) {
 	values := helper_get_params()
 	values[KindParam] = []string{"someKinda"}
 	values[NamespaceParam] = []string{"someNamespace"}
-	values[NameParam] = []string{"someName"}
+	values[NameParam] = []string{"someName.xx"}
```
kritikapradhan, line 94: is this test related to ur change ?

pkg/sloop/queries/respayloadquery.go
```diff
@@ -40,32 +46,40 @@ func GetResPayload(params url.Values, t typed.Tables, startTime time.Time, endTi
...
 		// get the previous key for those who has same payload but just before startTime
 		var getPreviousErr error
 		seekKey := getSeekKey(keyComparator, startTime)
+		glog.V(glogVerbose).Infof("GetResPayload: seekKey: %v", seekKey.String())
 		previousKey, getPreviousErr = t.WatchTable().GetPreviousKey(txn, seekKey, keyComparator)
 
 		// when getPreviousErr is not nil, we will not return err since it is ok we did not find previous key from startTime,
 		// we can continue using the result from rangeRead to proceed the rest payload
 		if getPreviousErr == nil {
+			glog.V(glogVerbose).Infof("GetResPayload: previousKey: %v", previousKey.String())
```
kritikapradhan, line 69: isn't previousKey the same as seekKey?

pkg/sloop/queries/respayloadquery.go
```diff
@@ -91,11 +105,11 @@ func GetResPayload(params url.Values, t typed.Tables, startTime time.Time, endTi
...
-//todo: add unit tests
 func getSeekKey(keyComparator *typed.WatchTableKey, startTime time.Time) *typed.WatchTableKey {
-	seekKey := keyComparator
-	seekKey.PartitionId = untyped.GetPartitionId(startTime)
-	seekKey.Timestamp = startTime
+	kind := keyComparator.Kind
+	namespace := keyComparator.Namespace
+	name := keyComparator.Name
+	seekKey := typed.NewWatchTableKey(untyped.GetPartitionId(startTime), kind, namespace, name, startTime)
 	return seekKey
```
kritikapradhan, line 113: NewWatchTableKey returns a pointer, right?

pkg/sloop/queries/respayloadquery.go
```diff
@@ -91,11 +105,11 @@ func GetResPayload(params url.Values, t typed.Tables, startTime time.Time, endTi
...
 
-//todo: add unit tests
 func getSeekKey(keyComparator *typed.WatchTableKey, startTime time.Time) *typed.WatchTableKey {
-	seekKey := keyComparator
-	seekKey.PartitionId = untyped.GetPartitionId(startTime)
-	seekKey.Timestamp = startTime
+	kind := keyComparator.Kind
+	namespace := keyComparator.Namespace
+	name := keyComparator.Name
+	seekKey := typed.NewWatchTableKey(untyped.GetPartitionId(startTime), kind, namespace, name, startTime)
```
kritikapradhan, line 112: NewWatchTableKey returns a pointer, right?

kritikapradhan:
Seems like u need to sign the salesforce-cla and tie it to ur commit.

pkg/sloop/queries/eventquery_test.go
duke-harlan: @kritikapradhan no, a test was failing, this fixed it

pkg/sloop/queries/respayloadquery.go
duke-harlan: it might be. seekKey is where to start looking. previousKey is the first key found.
Since seekKey is set with the timeStamp & partitionId based on the range read start time (based on where the user clicked), it seems unlikely

pkg/sloop/queries/respayloadquery.go
duke-harlan: yes, a pointer to a new instance

sana-jawad (approved):
Looks great! thanks @duke-harlan for fixing it

PR #173, opened 2021-09-09
Fix for "Could not get event record" error and cleaned up logging

sana-jawad:
1. Fix for error seen while ingesting data

Processing for updateEventCountTable failed with error Could not get event record: invalid key for table eventcount: /eventcount/001630076400/Pod/namespace/podname: key should have 6 parts: 

2. Also cleaned up logging and added a configuration to change the setting as needed.
3. Update debug page to show data from all tables.


pkg/sloop/processing/watch.go
```diff
@@ -124,3 +126,21 @@ func toWatchTableKeyPrefix(ts *timestamp.Timestamp, kind string, namespace strin
 
 	return typed.NewWatchTableKey(untyped.GetPartitionId(timestamp), kind, namespace, name, time.Time{}), nil
 }
+
+func GetUidForWatchEntry(tables typed.Tables, txn badgerwrap.Txn, kind string, namespace string, name string, timestamp time.Time) (string, error) {
+	watchKeyComparator := typed.NewWatchTableKeyComparator(kind, namespace, name, timestamp)
+	seekKey := queries.GetSeekKey(watchKeyComparator, time.Now())
```
akrag, line 132: Do we need a nil check here?

pkg/sloop/processing/eventcount.go
```diff
@@ -69,6 +70,15 @@ func updateEventCountTable(
 
 	eventCountByMinute := spreadOutEvents(computedFirstTs, computedLastTs, computedCount)
 
+	if involvedObject.Uid == "" {
```
akrag, line 73: can involvedObject be nil here? Are we receiving only involvedObject without Uid here?

pkg/sloop/processing/eventcount.go
```diff
@@ -69,6 +70,15 @@ func updateEventCountTable(
...
+		glog.V(common.GlogVerbose).Infof("Got empty Uid for name: %v, namespace: %v, kind: %v for event: %v ", involvedObject.Name, involvedObject.Namespace, involvedObject.Kind, newEventInfo.Reason)
+		returnedUid, err := GetUidForWatchEntry(tables, txn, involvedObject.Kind, involvedObject.Namespace, involvedObject.Name, time.Time{})
+		if err != nil {
+			return err
```
akrag, line 77: errors.Wrap(err, "Error retrieving Uid for the Watch Entry") ?

pkg/sloop/processing/eventcount.go
sana-jawad: No Its already checked here https://github.com/salesforce/sloop/blob/master/pkg/sloop/processing/processing.go#L56 Yes we are receiving some involved objects with Uid empty

pkg/sloop/processing/watch.go
sana-jawad: All the fields have already been checked in parent function

pkg/sloop/processing/eventcount.go
sana-jawad: sure! updated

PR #174, opened 2021-09-10
Cleaned up logs for Sloop

sana-jawad:
There were a lot of log lines in Sloop which were causing clogging of logs. Updated them to use verbosity and display when needed. There were also a lot of log lines published because of badger. Added a configuration to turn them on/off based on need. By default they are turned off.

### DIFF
--- a/pkg/sloop/queries/respayloadquery.go
+++ b/pkg/sloop/queries/respayloadquery.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"github.com/dgraph-io/badger/v2"
 	"github.com/golang/glog"
+	"github.com/salesforce/sloop/pkg/sloop/common"
 	"github.com/salesforce/sloop/pkg/sloop/kubeextractor"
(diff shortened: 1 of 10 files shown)


pkg/sloop/queries/respayloadquery.go
```diff
@@ -105,7 +102,7 @@ func GetResPayload(params url.Values, t typed.Tables, startTime time.Time, endTi
 	return bytes, nil
 }
 
-func getSeekKey(keyComparator *typed.WatchTableKey, startTime time.Time) *typed.WatchTableKey {
+func GetSeekKey(keyComparator *typed.WatchTableKey, startTime time.Time) *typed.WatchTableKey {
```
kritikapradhan, line 105: this change to global is for the other change right?

sana-jawad: yes moved it from this change. Thanks for pointing it out.

kritikapradhan (approved):

PR #175, opened 2021-09-10
Add debug option to view table related from all badger tables

sana-jawad:
In Debug -> Query Data Store page, there is an option to search only by particular table. In order to make debugging easier, adding an option to search in all the tables and give a complete picture which can help in debugging different issues. Also in the same page the encoded value used to update the text box so fixed that as well.


pkg/sloop/webserver/debug.go
```diff
@@ -124,32 +124,44 @@ func listKeysHandler(tables typed.Tables) http.HandlerFunc {
...
-				totalCount++
-				thisKey := string(itr.Item().Key())
-				if keyRegEx.MatchString(thisKey) {
-					keys = append(keys, thisKey)
-					count += 1
-					totalSize += itr.Item().EstimatedSize()
-					if count >= maxRows {
-						glog.Infof("Number of rows : %v has reached max rows: %v", count, maxRows)
-						break
+			for _, tablename := range tablesToSearch {
```
kritikapradhan, line 137: can this be tested?

sana-jawad: We have not been testing debug pages as they are purely for debug purpose and not customer facing.

pkg/sloop/webserver/webfiles/filter.js
```diff
@@ -16,7 +16,7 @@ function getUrlVars() {
 function getUrlParam(parameter, defaultvalue){
     var urlparameter = defaultvalue;
     if(window.location.href.indexOf(parameter) > -1){
-        urlparameter = getUrlVars()[parameter];
+        urlparameter = decodeURIComponent(getUrlVars()[parameter]);
```
kritikapradhan, line 19: Is this related to table change?

sana-jawad: Yes its for the same page. The text box in the page gets encoded every time we submit the page

pkg/sloop/webserver/debug.go
kritikapradhan: :( 
Should L137 be `for tablename:= range tablesSearch {` ?

kritikapradhan (approved):

PR #176, opened 2021-09-10
Fixed missing UID bug

sana-jawad:
There was an error observed at times while ingesting event count:

`Processing for updateEventCountTable failed with error Could not get event record: invalid key for table eventcount: /eventcount/001630076400/Pod/namespace/podname: key should have 6 parts:`

Fix:
For a certain type of events from Tain Manager the involved object in Event body didn't have UID.
Previously we used to drop all these events and they didn't show up in Sloop UI.
Now, we find the record in Watch table and retrieve the UID if dont get it in Event payload. Hence these payloads are ingested now. 
Also added unit test cases for testing of this scenario.


akrag (approved):
LGTM

mengyaoyang11 (approved):
Approve

PR #177, opened 2021-09-10
No events bug fix

sana-jawad:
After a recent change the events stopped showing up in Sloop UI when there was an event indicated by yellow/green capsules.

    Fix:

    In event type we have names in watch key as /watch/partitionID/Event/Namespace/Name.xx

    With the recent change it resulted in watch key to be /watch/partitionID/Event/Namespace/Name/ and thus no events matched this prefix. This a fix to deal with Events type and make it work for all K8s kinds.

    Also added unit test cases and fixed the integration unit test case which should have caught the issue in the previous check in.


kritikapradhan:
LGTM

kritikapradhan (approved):

kritikapradhan (approved):

PR #181, opened 2021-09-25
Update README.md

annelausf:
Checking that GO111MODULE="auto" helps prevent the error that package 'github.com/dgraph-io/badger/v2' cannot be found.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,13 +53,12 @@ mkdir -p $GOPATH/src/github.com/salesforce
 cd $GOPATH/src/github.com/salesforce
 git clone https://github.com/salesforce/sloop.git
 cd sloop
+go env -w GO111MODULE=auto
 make
 $GOPATH/bin/sloop
 ```
 
 When complete, you should have a running Sloop version accessing the current context from your kubeConfig. Just point your browser at http://localhost:8080/
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,6 +59,8 @@ $GOPATH/bin/sloop
 
 When complete, you should have a running Sloop version accessing the current context from your kubeConfig. Just point your browser at http://localhost:8080/
 
+**NOTE**: Before running ```make```, run ```go env``` and check that the module environment variable is set to ```GO111MODULE=auto```. If it is not run ```go env -w GO111MODULE=auto``` and then ```make```.
```
kritikapradhan, line 62: You can also think about editing the sh code(L51-57) to the following -
 ```sh
mkdir -p $GOPATH/src/github.com/salesforce
cd $GOPATH/src/github.com/salesforce
git clone https://github.com/salesforce/sloop.git
cd sloop
go env -w GO111MODULE=auto
make
$GOPATH/bin/sloop
```

annelausf: If I were to change it - should I include context on why that line is included as a comment beside it?

kritikapradhan: not required, IMO

kritikapradhan (approved):

PR #186, opened 2021-10-19
Added indication when there are no payload / events for a given period

annelausf:
Table in the Details pop-up does not convey when there are no payloads / events, users could assume the events are loading. 

Before:
Screen Shot 2021-10-18 at 5 17 03 PM

Fix:
Screen Shot 2021-10-18 at 5 18 01 PM


sana-jawad (approved):
Looks good to me!

Issue #157, opened 2021-05-25
Error in initializing K8s client: no Auth Provider found for name "azure"


sumitjainn:
Hi,

I am getting the below error when trying to run sloop in a docker container

```
kubeclient.go:42] Creating k8sclient with user-defined config masterURL=, kubeContext=*****.
mysloop_1  | E0525 17:05:16.909310       1 kubeclient.go:53] Cannot Initialize Kubernetes Client API: no Auth Provider found for name "azure"
mysloop_1  | I0525 17:05:16.910185       1 store.go:112] Closing store
mysloop_1  | badger 2021/05/25 17:05:16 INFO: Got compaction priority: {level:0 score:1.73 dropPrefix:[]}
mysloop_1  | I0525 17:05:16.918293       1 store.go:114] Finished closing store
mysloop_1  | E0525 17:05:16.934304       1 main.go:31] Main exited with error: failed to create kubernetes client: no Auth Provider found for name "azure"

```

Looks like https://github.com/salesforce/sloop/blob/master/pkg/sloop/ingress/kubeclient.go does not import the azure auth plugin as mentioned in https://github.com/liggitt/kubernetes/blob/master/staging/src/k8s.io/client-go/examples/README.md#auth-plugins

`import _ "k8s.io/client-go/plugin/pkg/client/auth/azure"`

sana-jawad:
thanks @sumitjainn for reporting this issue. We have added it in our list of work items and we will update this issue once we have an update. Please feel free to contribute as well.

nikhilgorantla:
any update on this ?

sumitjainn:
@sana-jawad Please review the PR https://github.com/salesforce/sloop/pull/191 for fixing this issue.
Also, signing CLA failed for me, not sure where the problem is.

sana-jawad:
@sumitjainn I have approved the PR and merged it. Thanks for the contribution.

PR #191, opened 2021-11-05
Fix for #157: Load Azure auth plugin for K8s client

sumitjainn:
Fixes the issue https://github.com/salesforce/sloop/issues/157

Load the azure auth plugin to prevent initialization failure in case k8s clusters use azure authentication.

Reproduced and fix validated on local. 

@sana-jawad Please review


sana-jawad (approved):
Thanks @sumitjainn for the contribution

PR #192, opened 2021-11-09
add handling for special delete object

duke-harlan:
P: When receiving cache.DeletedFinalStateUnknown in reportDelete(), the object is contained inside. Downstream processing wasn't recognizing it as a k8s object.
S: Process the contained object as the delete instead of the cache.DeletedFinalStateUnknown object

I believe all the changes to bindata.go are formatting changes.

### DIFF
--- a/pkg/sloop/ingress/kubewatcher.go
+++ b/pkg/sloop/ingress/kubewatcher.go
@@ -270,6 +270,11 @@ func (i *kubeWatcherImpl) reportAdd(kind string) func(interface{}) {
 
 func (i *kubeWatcherImpl) reportDelete(kind string) func(interface{}) {
 	return func(obj interface{}) {
+		delObj, ok := obj.(cache.DeletedFinalStateUnknown)
+		if ok {
+			obj = delObj.Obj
+		}
+
 		watchResultShell := &typed.KubeWatchResult{
 			Timestamp: ptypes.TimestampNow(),
 			Kind:      kind,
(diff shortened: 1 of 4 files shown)


pkg/sloop/ingress/kubewatcher.go
```diff
@@ -270,6 +270,11 @@ func (i *kubeWatcherImpl) reportAdd(kind string) func(interface{}) {
 
 func (i *kubeWatcherImpl) reportDelete(kind string) func(interface{}) {
 	return func(obj interface{}) {
+		delObj, ok := obj.(cache.DeletedFinalStateUnknown)
```
sana-jawad, line 273: Is there a chance that delete object is in the format that other Add or Update objects are @duke-harlan ?

duke-harlan: @sana-jawad the object passed in could me any k8s object. This cast returns ok = true if the object is of that type, so the only time this has any affect is when it is a DeletedFinalStateUnknown. Otherwise it passes through as before.

pkg/sloop/processing/processing.go
```diff
@@ -55,6 +55,7 @@ func (r *Runner) Start() {
 			if err != nil {
 				r.processingFailed("cannot extract resource metadata", err)
 			}
+			glog.V(99).Infof("watchRec metadata: %v", resourceMetadata)
```
sana-jawad, line 58: I would suggest to add loglevel 99 in the constants file and use that. We already have it for other levels

pkg/sloop/webserver/bindata.go
```diff
@@ -413,20 +413,20 @@ func AssetNames() []string {
 
 // _bindata is a table, holding each asset generator, mapped to its name.
 var _bindata = map[string]func() (*asset, error){
-	"webfiles/debug.html": webfilesDebugHtml,
```
sana-jawad, line 416: I think we only need binddata update if there is any ui file update.

sana-jawad (approved):

Issue #194, opened 2021-11-25
Time parsing and timezone issues


triqster:
There seems to be an issue with the date and time parsing, resulting in broken queries and therefor an empty GUI result.

Steps to reproduce:

- In the GUI enter a date and time like: 
![image](https://user-images.githubusercontent.com/5626877/143422630-56174f22-3809-48ab-838c-ccb421622537.png)

The important part here is to end with `00`

```
E1125 10:06:56.445817       1 query.go:42] computeTimeRange failed with error: parsing time "2021-11-25T08:52" as "2006-01-02T15:04:05": cannot parse "" as ":"
```
I'm  also constantly running into a already reported bug with little to no progression it seems, the query is in UTC while the "now" button and the shown time is in my current time zone (UTC +1) so I'm constantly getting errors like:
![image](https://user-images.githubusercontent.com/5626877/143422999-c97119ec-fb5c-40a0-aeac-5d63d089c1de.png)

mengyaoyang11:
Hi @triqster thanks for reporting the issue and sorry for the late response. Could you please take a look at this discussion which reported the same issue with yours https://github.com/salesforce/sloop/issues/167. i would recommend you check your browser, ideally the date time picker should have AM/PM option for you to pick. I would recommend switch web browser or update the chrome version. The chrome version I use is 92.0.4515.131, can be used as a reference. Thanks, let me know how that works.

mengyaoyang11:
The parsing issue you reported I attached a screenshot, its not happening on my side, again, could you please check your browser and see why the AM/PM option is missing. BTW, Some customer reported the am/pm option works only with firefox

![Screen Shot 2021-12-20 at 5 59 13 PM](https://user-images.githubusercontent.com/78186416/146857962-830e4f89-8108-4efd-b7a8-c1c435a500ea.png)

Issue #167, opened 2021-08-03
Can not submit search query


xcrezd:
After clicking on the "Now" button and then "submit" got the error 

![image](https://user-images.githubusercontent.com/1153266/128002882-67b1156f-89be-4a8b-98df-1b32c717aa8b.png)

mengyaoyang11:
Hi @xcrezd, thanks for submitting the issue, could you please specify how/where you run sloop, if possible, could you please send me the link of the sloop instance you were running? Also, did you rebase with master to get the latest code? We appreciate if you can provide more detail which is needed to investigate the issue. Thanks!

mengyaoyang11:
Also, i am wondering if you can choose the AM or PM through the date time picker, in your UI I didn't see it, ideally, it should look like this
<img width="449" alt="Screen Shot 2021-08-03 at 1 25 14 PM" src="https://user-images.githubusercontent.com/78186416/128081433-7e5ef778-bfaa-431f-8de9-6a5e2bc585f4.png">

<img width="285" alt="Screen Shot 2021-08-03 at 1 23 27 PM" src="https://user-images.githubusercontent.com/78186416/128081155-3724da61-0515-48b3-a96a-9aedc85d0e2a.png">

BTW, the `Now` button renders the current time in UTC format, so it's possible the error pops up if you select any time later than UTC now from date time picker.

xcrezd:
You are right, I miss the AM/PM option.

I have built the docker image from the commit 
```
e86b22215520900bf060e92c60c67da9ef97ef6c (HEAD -> master, tag: v1.1, origin/master, origin/HEAD)
Author: mengyaoyang11 <78186416+mengyaoyang11@users.noreply.github.com>
Date:   Wed Jun 16 16:04:25 2021 -0700
    Present User Selected Time and Now in UTC format (#161)
```

This is the last one until now.

Also, I have these errors and messages in the Chrome developer console 
![image](https://user-images.githubusercontent.com/1153266/128178204-b2172dba-f37a-4363-8333-de61f043baf8.png)

Sloop is running on EKS 1.17v cluster, the web access is via 
`kubectl port-forward -n kube-system sloop-0 8080:8080`


Hope it's more informative.

Thanks ahead, 
Dima

mengyaoyang11:
Can you try building sloop from source, instruction is here: https://github.com/salesforce/sloop#build-from-source and let me know how that works? Thanks

xcrezd:
[...]

PR #201, opened 2022-02-22
Remove selectedEndTime.max

ryanbrainard:
As submitted in https://github.com/salesforce/sloop/issues/194 and https://github.com/salesforce/sloop/issues/167, users are running into issues with the max end time. I ran into this too and it makes the app unusable. I looked into this, and there are three issues:

- `iso` is adjusted from `utcNow`, but then calling `toISOString()`, so essentially double offset
- the `elem.max` is only set on `$(document).ready`, which means if a user presses the "Now" button, it ends up after the max
- `elem.value` is set twice

I started to fix these, but stepping back, I don't understand why we need a max at all. It really should be just "now" when the server accepts the request, not when the page first (or even subsequently) loaded on the client side. As such, this drops the setting of `elem.max`. It also cleaned up the unused vars and the double setting of `elem.value`.

Fixes https://github.com/salesforce/sloop/issues/167
Fixes https://github.com/salesforce/sloop/issues/194
cc: @mengyaoyang11


mengyaoyang11:
@ryanbrainard thanks for the contribution! could you please attach screenshot or describe the effect  after your change(removing maxTime)? It would be great if we have the final fix before we approve/merge anything, only removing maxTime wouldn't fix anything, right?

ryanbrainard:
@mengyaoyang11 Thanks for the quick response. 

> could you please attach screenshot or describe the effect after your change(removing maxTime)?

The effect is that a max end time is no longer enforced. This means that the error reported in https://github.com/salesforce/sloop/issues/194 no longer happens.  There's not really anything to screenshot since it's removing the error.

Note, I believe the reason there was difficulty consistently reproducing https://github.com/salesforce/sloop/issues/194 and https://github.com/salesforce/sloop/issues/167 is because of timezones. For users in the western hemisphere (UTC-), the max would have been in the future so they would not see the problem, but for users (like me and @xcrezd) in the eastern hemisphere (UTC+) the max would be in the past causing the problem reported.

As I mentioned above, while we could fix the timezone offset problem, having a max time isn't really needed, and even if we did, it doesn't make sense to do on the client side. If a user chooses a time in the future, then the query should still work. In fact this could be helpful to do on purpose to avoid having to click "Now" to get the latest data.

> It would be great if we have the final fix before we approve/merge anything, only removing maxTime wouldn't fix the issue described, right?

This is the final fix and should address the issue.

ryanbrainard:
I noticed the [Updating webfiles folder](https://github.com/salesforce/sloop#updating-webfiles-folder) section of the readme that says to run `go-bindata`. I tried doing this with both `github.com/jteeuwen/go-bindata` (from `go.mod`; however is marked as `v3.0.7+incompatible`) and `github.com/go-bindata/go-bindata`, but both of them produce pretty different diffs from what is currently in this repo. Which fork and version of `go-bindata` should I use?

`github.com/go-bindata/go-bindata` seems to be closer to what's on `master`, so going to push up that, but it would be good to have this documented in the readme and have it match what it is used in `go.mod`.

cc: @annelausf (last edit to `pkg/sloop/webserver/bindata.go`)
cc: @kritikapradhan, @tyrken (readme instructions)

mengyaoyang11:
@ryanbrainard got it, thanks for the explanation, ran locally and the fix removed maxTime totally, so I am okay to approve/merge your changes, Thanks!

For updating bindata.go file, I think all you need to do is to run `go-bindata -pkg webserver -o bindata.go webfiles/ ` within webserver directory, are you running the command under the correct directory? this [commit](https://github.com/salesforce/sloop/pull/201/commits/31a6ad68668ee276b66f9e8bf3bcb658ac776731) seems doing the right work, any other concerns?

ryanbrainard:
> For updating bindata.go file, I think all you need to do is to run go-bindata -pkg webserver -o bindata.go webfiles/ within webserver directory, are you running the command under the correct directory? this commit seems doing the right work, any other concerns?

Yes, that's what I did, and it does seem ok for this PR. My concern was that there are multiple forks and versions of go-bindata all with different formats, so wanted to make I was using the right one.

kritikapradhan:
@ryanbrainard thanks for looking into this. 

Context for the feature: To reiterate that we do not store data longer than 2 weeks, as a result of which that avoids allowing customers to even select a range outside of that. Having said that catering to different time zones is something we will need to look into. 

Wrt to the go-bindata version, `github.com/go-bindata/go-bindata ` is correct. Thanks for pointing out that go.mod should reflect that.

sana-jawad (approved):

ryanbrainard:
> Context for the feature: To reiterate that we do not store data longer than 2 weeks, as a result of which that avoids allowing customers to even select a range outside of that. Having said that catering to different time zones is something we will need to look into.

Thanks, that help; however, in that case, wouldn't we want a _min_ end time, not a max end time?

> If it is possible to account for the above context in your changes, that would be great.

While we could move it to a `min` 2 weeks in the past offset by the time range, doing this on the client side is still going to have the problem of the date getting out of sync with clock time since this is set on page load, not to mention problems with the changes to the offset. In my opinion, it would be better to remove the validation for now (this PR), and then we can always add a more robust solution if validation is still needed. This would at least make the app usable for anyone in UTC+ timezones.

ryanbrainard:
> Wrt to the go-bindata version, `github.com/go-bindata/go-bindata ` is correct. Thanks for pointing out that go.mod should reflect that.

Logged https://github.com/salesforce/sloop/issues/203 for this.

sana-jawad:
> > Context for the feature: To reiterate that we do not store data longer than 2 weeks, as a result of which that avoids allowing customers to even select a range outside of that. Having said that catering to different time zones is something we will need to look into.
> 
> Thanks, that help; however, in that case, wouldn't we want a _min_ end time, not a max end time?
> 
> > If it is possible to account for the above context in your changes, that would be great.
> 
> While we could move it to a `min` 2 weeks in the past offset by the time range, doing this on the client side is still going to have the problem of the date getting out of sync with clock time since this is set on page load, not to mention problems with the changes to the offset. In my opinion, it would be better to remove the validation for now (this PR), and then we can always add a more robust solution if validation is still needed. This would at least make the app usable for anyone in UTC+ timezones.

I agree @ryanbrainard

PR #212, opened 2022-04-19
Added some more platforms and tags to the package

sana-jawad:


.github/workflows/docker.yml
```diff
@@ -24,21 +24,27 @@ jobs:
         uses: docker/metadata-action@v3
         with:
           images: ghcr.io/${{ github.repository }}
+          # generate Docker tags based on the following events/attributes
+          tags: |
```
duke-harlan, line 28: is it ok to have multiple tags with the same name & different values?

duke-harlan (approved):
a question, but otherwise lgtm

.github/workflows/docker.yml
sana-jawad: @duke-harlan I followed an example online. After check in another package will be published, I will answer this question after reviewing its tags.

PR #218, opened 2022-06-16
External chain static resource localization.Fix the template applicat…

zillaer:
External chain static resource localization.Fix the template application failure


zillaer:
salesforce-cla sign up failure

sana-jawad:
@zillaer Can you please provide some more context of what issue this PR is solving?

PR #226, opened 2022-08-10
W-11339557 fix end date time which comes with 0 milliseconds

sarjamil:


pkg/sloop/webserver/webfiles/filter.js
```diff
@@ -108,6 +108,10 @@ function setFiltersAndReturnQueryUrl(defaultLookback, defaultKind, defaultNamesp
         var now =  new Date().toISOString()
         selectedEndTime = now.substring(0, now.length-5)
     }
+    else{
```
sana-jawad, line 111: nit: space between else and {

pkg/sloop/webserver/webfiles/filter.js
```diff
@@ -108,6 +108,10 @@ function setFiltersAndReturnQueryUrl(defaultLookback, defaultKind, defaultNamesp
...
+        var endTime = new Date(selectedEndTime);
+        selectedEndTime = endTime.toISOString().slice(0,-1)
```
sana-jawad, line 113: please make sure to test a bunch of use cases to make sure there is no regression

sana-jawad (approved):

pkg/sloop/webserver/webfiles/filter.js
sarjamil: good idea, i added two cases for end date in time range test

PR #242, opened 2022-12-13
Fix publish GHA CI - Arm32 doesn't support Go race detector

kskewes-sf:
We still use `go test -race` in the `test.yml` GHA so will catch failures there.

Whilst in the area, change README from Travis to GHA based CI status.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,4 @@
-name: Format, lint and test
+name: Test
 
 on:
   push:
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -1,4 +1,4 @@
-name: Format, lint and test
+name: Test
```
venkatramreddykunta, line 1: +1. 
Test makes sense.

venkatramreddykunta (approved):
LGTM

venkatramreddykunta (approved):

.github/workflows/test.yml
kskewes-sf: True. I had simplified it down for the GitHub Status Badges added to README.md.
As part of #245 might change again.

Issue #210, opened 2022-03-28
Relax probing for forbidden resources


hoerup:
Useing latest sloop on a 1.20.x cluster

In the log it seems that sloop continues to query for resources that it doesn't have access to

IMHO, when sloop has detected a forbidden - it should mark that resourcetype as not accessible and back off
(it could then perhaps retry at a later point eg every CrdRefreshInterval * 4 )

E0328 10:40:10.785774       1 reflector.go:156] pkg/mod/k8s.io/client-go@v0.17.0/tools/cache/reflector.go:108: Failed to list *unstructured.Unstructured: servicemonitors.monitoring.coreos.com is forbidden: User "system:serviceaccount:sloop:sloop" cannot list resource "servicemonitors" in API group "monitoring.coreos.com" at the cluster scope
I0328 10:40:10.986713       1 trace.go:116] Trace[320645588]: "Reflector ListAndWatch" name:pkg/mod/k8s.io/client-go@v0.17.0/tools/cache/reflector.go:108 (started: 2022-03-28 10:39:58.587890641 +0000 UTC m=+1821.020996664) (total time: 12.398775147s):

$kubectl logs --since=30m -n sloop sloop-0 | grep servicemonitors | wc -l
134
so it queries for the same forbidden resource type aproximately 4.47 times pr minute

hoerup:
Or maybe event better: add config options for positivelist or negativelist resourcegroups to monitor

Issue #206, opened 2022-03-28
Namespace selectors


hoerup:
Please add config options for configuring which namespaces to watch for changes, preferable one for excluding some namespace and one for include-only

PR #259, opened 2023-04-27
Add exclusion rules to filter unwanted events

acbox:


hoerup:
could possibly solve #206 ?

hoerup:
can the filtering also exclude CRD's ?
and thereby solve  #210

Tiltfile
```diff
@@ -0,0 +1,57 @@
+#
```
nurland, line 1: Can you please move this file out to a separate PR, we'd like to keep PRs focused on one feature, thanks!
It would also be great if we can update readme with instructions on how to run tile with sloop.

pkg/sloop/common/utilities.go
```diff
@@ -37,3 +37,13 @@ func Contains(stringList []string, elem string) bool {
 func GetFilePath(filePath string, fileName string) string {
 	return path.Join(filePath, fileName)
 }
+
+func Truncate(text string, width int) (string, error) {
```
nurland, line 41: can you please add unit tests for Truncate?

pkg/sloop/ingress/kubewatcher_test.go
```diff
@@ -81,21 +82,58 @@ func Test_bigPicture(t *testing.T) {
...
+	}
+
+	// await events
+	result1 := <-outChan
+	result2 := <-outChan
+	result3 := <-outChan
+
+	if !strings.Contains(result1.Payload, `"name":"ns"`) { t.Error("Expected event not found") }
+	if !strings.Contains(result2.Payload, `"name":"s1"`) { t.Error("Expected event not found") }
+	if !strings.Contains(result3.Payload, `"name":"s3"`) { t.Error("Expected event not found") }
```
nurland, line 136: should we check that `"name":"s2"` is not present?

pkg/sloop/ingress/kubewatcher_test.go
```diff
@@ -81,21 +82,58 @@ func Test_bigPicture(t *testing.T) {
 	masterURL := "url"
 	kubeContext := "" // empty string makes things work
 	enableGranularMetrics := true
-	kw, err := NewKubeWatcherSource(kubeClient, outChan, resync, includeCrds, time.Duration(10*time.Second), masterURL, kubeContext, enableGranularMetrics)
+	exclusionRules := map[string][]any{
```
nurland, line 85: can we also add a test where exclusionRules are not provided, i.e. uses the default value of `ExclusionRules:           map[string][]any{},`

nurland (commented):
Thank you @acbox for the contribution, this is great! Added some comments, but otherwise lgtm

pkg/sloop/ingress/kubewatcher_test.go
acbox: I've added seperate tests for default and non-default exclusion rules

Tiltfile
acbox: Removed, not sure this warrants a separate PR yet as there were some unresolved issues with the workflow incorporating Tilt. The Tiltfile also relied on running a local cluster with Kind, which is perhaps too opinionated.

pkg/sloop/common/utilities.go
acbox: Added unit tests, doing this highlighted some bugs in Truncate that I've fixed.

pkg/sloop/ingress/kubewatcher_test.go
acbox: Could you give an example? I found this the most straightforward way. I changed the `if` statements to `assert` calls, bit cleaner.

pkg/sloop/ingress/kubewatcher.go
```diff
@@ -360,6 +373,41 @@ func (i *kubeWatcherImpl) refreshCrdInformers(masterURL string, kubeContext stri
...
+func (i *kubeWatcherImpl) eventExcluded(kind string, resourceJson string) (bool) {
+	filters := i.getExclusionRules(kind)
+	for _, logic := range filters {
+		logicJson, err := json.Marshal(logic)
+		if err != nil {
+			glog.Errorf(`Failed to parse event filtering rule "%s": %s`, logic, err)
+			return false
+		}
+		var result bytes.Buffer
+		err = jsonlogic.Apply(
```
nurland, line 396: please add err check

acbox: Done

nurland (approved):
Thank you for your contribution! LGTM

acbox:
Pushed a fix for the [failed](https://github.com/salesforce/sloop/actions/runs/4927800936/jobs/8805861492) GitHub Action test and signed the CLA.

nurland (approved):

acbox:
This PR should pass the CI tests now as per the commit comment here https://github.com/salesforce/sloop/pull/259/commits/ae0b4d0a403078f4cad62948bda67746d8ac3ded

sana-jawad:
@acbox thanks for your contribution.

PR #262, opened 2023-05-18
Add GHA workflow to lint helm chart, and fix a bug in the chart

dlipovetsky:
1. Valid apiGroup values, e.g. `*.mygroup.example.com`, must be quoted in the template.

    Before this change, `helm lint` failed with this error:

    `[ERROR] templates/clusterrole.yaml: unable to parse YAML: error converting YAML to JSON: yaml: line 92: did not find expected alphabetic or numeric character`

2. Adds an action to lint the helm chart whenever it is changed.

    Note: I've fixed the helm version at 3.12.0. If you'd like to always use the latest stable version, the action will need to use a GitHub token, as explained here: https://github.com/Azure/setup-helm#example

### DIFF
--- a/.github/workflows/lint-helm-chart.yml
+++ b/.github/workflows/lint-helm-chart.yml
@@ -1,0 +1,32 @@
+name: Lint helm chart
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+          - 'helm/sloop/**'
+  pull_request:
+    branches:
+      - "*"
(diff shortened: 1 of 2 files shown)


dlipovetsky:
I signed the CLA, but it looks like the service had some internal error.

sana-jawad:
Looks good to me. Thanks for the contribution @dlipovetsky

PR #264, opened 2023-05-31
feat: Compress data backup using zstd

dlipovetsky:
The backup turns out to be highly compressible!

I chose zstd. Although gzip is standard, zstd is faster, and compresses better.

Example:

```
4.6M sloop-cluster-0.bak.zst
9.5M sloop-cluster-0.bak.gz
155M sloop-cluster-0.bak
```

This introduces a dependency on module github.com/klauspost/compress, which contains a pure Go zstd implementation, so it can be used whether sloop is built with CGO enabled or disabled.


annelau21 (approved):
Looks good to me

PR #284, opened 2026-02-13
@W-21199400:adding-log-levels

shardulnegi-boop:
This change introduces explicit log verbosity levels to sloop to mitigate log-flooding issues. Currently, certain events—specifically date/time parsing failures—are being logged at a high frequency. This constant posting to the logs creates significant memory pressure and makes it difficult to monitor actual system health.

By moving these specific logs to LogLevelDebug (V2), we reduce the overhead on the logging subsystem while still allowing for deep troubleshooting when needed.

### DIFF
--- a/pkg/sloop/kubeextractor/events.go
+++ b/pkg/sloop/kubeextractor/events.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/golang/glog"
+	"github.com/salesforce/sloop/pkg/sloop/common"
 )
 
 // Example Event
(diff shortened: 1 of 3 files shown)


pkg/sloop/kubeextractor/events.go
```diff
@@ -91,14 +92,14 @@ func ExtractEventInfo(payload string) (*EventInfo, error) {
 
 	fs, err := time.Parse(time.RFC3339, internalResource.FirstTimestamp)
 	if err != nil {
-		glog.Errorf("Could not parse first timestamp %v\n", internalResource.FirstTimestamp)
+		glog.V(common.LogLevelDebug).Infof("Could not parse first timestamp %v", internalResource.FirstTimestamp)
```
osassonSF, line 95: Any concern we should leave this log to investigate why it keeps happening?

osassonSF: What's the benefit of removing this log?

osassonSF (approved):

PR #290, opened 2026-06-10
prevent startup deadlock when watching many CRDs

RezaMash:
On clusters with a large number of CRDs (e.g. 500), sloop would
hang during startup and never bind its webserver on :8080. Liveness and
startup probes against /healthz then got "connection refused" forever,
putting the pod into a permanent CrashLoopBackOff.

Root cause was a lock-ordering deadlock during the initial informer sync:

  - writeToOutChan() sent to the bounded kubeWatchChan (buffer 1000) while
    holding i.protection. The author had already flagged this line as
    dangerous.
  - The single processing goroutine drains that channel doing several
    synchronous, fsync'd (SyncWrites=true) BadgerDB transactions per event,
    so it could not keep up with the initial-sync burst from the CRD
    informers plus the core resources.
  - Once the channel filled, informer handlers blocked on the send while
    holding i.protection. Meanwhile the main goroutine was still in
    startCustomInformers() iterating over every CRD, and each informer
    setup also needs i.protection. It blocked, so NewKubeWatcherSource()
    never returned and webserver.Run() was never reached.

Two changes:

1. ingress/kubewatcher.go: do not hold i.protection across the channel
   send. Take the lock only to read `stopped`, release it, then send via a
   select on i.outchan / i.stopChan. The send no longer blocks the lock,
   and it unblocks promptly on shutdown instead of blocking forever on a
   full channel. This removes the deadlock.

2. server/server.go: construct the kube watcher in a background goroutine
   so the main goroutine reaches webserver.Run() and binds /healthz
   immediately, regardless of how long the initial CRD sync takes. This is
   defense in depth: even a merely slow (not deadlocked) watcher no longer
   causes probes to kill the pod during startup. Shutdown reads
   kubeWatcherSource under a mutex since it is now set from a goroutine.

With these changes the pod becomes Ready in seconds with watchCrds=true,
and CRD-backed resources are recorded.

Verified: go build ./pkg/..., go test ./pkg/sloop/ingress/...
./pkg/sloop/server/..., and go test -race ./pkg/sloop/ingress/... all pass.


RezaMash:
I signed the CLA.

RezaMash:
The PR is tagged with [cla:missing](https://github.com/salesforce/sloop/issues?q=state%3Aopen%20label%3Acla%3Amissing) but I sigend the CLA yesterday and when I try to sign it I get the following message:
image

